PR #8, opened 2025-08-18
fix: clone url

dschmidt:


orgaralf:
outdated and fixed in later PRs

Issue #7, opened 2025-08-18
[bug] Enable strict Content-Security-Policy headers


dschmidt:
Currently strict CSP cannot be enabled.

--- 
These are no-gos for our use case
```
    <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/js/bootstrap.bundle.min.js"></script>
    <script async data-uid="3a02cc4dec" src="https://synaplan.kit.com/3a02cc4dec/index.js"></script>
```

and also assuming our application is always running on `synplan.com` is not going to work out.

For that reason we should implement strict Content-Security-Policy headers, so we find current issues and avoid introducing new ones.

PR #47, opened 2025-08-25
feat: switch to frankenphp and enable xdebug

dschmidt:
I needed to setup debugging and thought "why not give frankenphp a shot while at it" ...

What do you think of it in principle?
Which of those headers were actually needed?
c.f. #7

### DIFF
--- a/dev/docker/Dockerfile
+++ b/dev/docker/Dockerfile
@@ -1,5 +1,5 @@
-# Use PHP 8.3 with Apache
-FROM php:8.3-apache
+# Use FrankenPHP with PHP 8.4
+FROM dunglas/frankenphp:php8.4
 
 # Install system dependencies including FFmpeg and whisper.cpp dependencies
 RUN apt-get update && apt-get install -y \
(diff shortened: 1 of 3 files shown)


orgaralf:
I am fine with FrankenPHP, but IMHO that is technical research and nice to have. More pressing issues are missing capabilities for companies to use synaplan as a drop-in replacement for ChatGPT (or as a gateway to it). We have a lot of capabilites that are not on par with Google Gemini or ChatGPT and that shold work ("upload and translate this PDF", eg.). We have the dilemma of "cool and faster PHP" with "features the users need". Currently I would say: FrankenPHP yes, but it does not matter for me personally. I am still developping directly on WSL with Apache and it should work on both... I will focus on the features and the robustness of the app, you can make it elegant :-)

dev/docker/Dockerfile
```diff
@@ -67,78 +67,24 @@ RUN cd /tmp && \
...
-    && echo 'Header always set Referrer-Policy "strict-origin-when-cross-origin"' >> /etc/apache2/conf-available/secure-basics.conf \
-    && echo 'Header always set Permissions-Policy "camera=(),microphone=(),geolocation=()"' >> /etc/apache2/conf-available/secure-basics.conf \
-    && a2enconf secure-basics
-
-# CSP Report-Only für DEV
-RUN echo 'Header always set Content-Security-Policy-Report-Only "default-src '"'"'self'"'"'; script-src '"'"'self'"'"' '"'"'unsafe-inline'"'"'; style-src '"'"'self'"'"' '"'"'unsafe-inline'"'"'; img-src '"'"'self'"'"' data: blob:; font-src '"'"'self'"'"' data:; connect-src '"'"'self'"'"' wss: https://api.openai.com https://api.anthropic.com https://generativelanguage.googleapis.com https://api.thehive.ai https://api.search.brave.com https://graph.search.brave.com https://graph.facebook.com; worker-src '"'"'self'"'"' blob:; media-src '"'"'self'"'"' blob:; manifest-src '"'"'self'"'"'; frame-src '"'"'self'"'"'; object-src '"'"'none'"'"'; base-uri '"'"'self'"'"'; form-action '"'"'self'"'"'; frame-ancestors '"'"'none'"'"'"' > /etc/apache2/conf-available/csp-reportonly.conf \
```
orgaralf, line 136: The app should be allowed to connect to all APIs outside the network by default. If it does OR not, is up to the sysadmin of the target system. You can use ChatGPT or Gemini, if you like, but you do not have to and just use ollama locally. If that is given, I am fine!

docker-compose.yml
```diff
@@ -20,6 +22,8 @@ services:
       - DB_USER=synaplan_user
       - DB_PASSWORD=synaplan_password
       - OLLAMA_SERVER=ollama:11434
+      - XDEBUG_MODE=debug
```
orgaralf, line 25: Yes, much needed... should also be a flag to disable for llive deployments?

dev/docker/Dockerfile
dschmidt: Does the frontend or the backend talk to those apis?

docker-compose.yml
dschmidt: Yes and no, the docker-compose.yml in the project root is just the dev stack - it's not prepared for production deployments in any way. We might or might not use the same docker-compose file. I would leave that for a follow up: #48

PR #52, opened 2025-08-27
feat: oidc

dschmidt:
includes #47


orgaralf:
Were you able to test it with a running keycloak? I am missing a local install to test it thouroughly...

PR #54, opened 2025-08-27
feat: ci

dschmidt:


yusuf57rl (approved):
Looks good 👍
Removing the extra checks keeps things simpler and more reliable. The volumes still handle caching, so no downsides.
Overall the changes make the setup cleaner and easier to follow.

PR #55, opened 2025-08-30
Add initial implementation of Again functionality with central merge/stream/reset handling

yusuf57rl:
- Introduced new Again feature
- Implemented central merge, stream, and reset logic for consistent handling
- Integrated Again responses to be properly stored and streamed
- Known issue: in image generation via Again, the original input text is still stored instead of the out-payload (to be fixed next)


orgaralf:
Ok, thanks <3... will merge now and work on the bugfix, so I can also do some refactoring on that.

PR #76, opened 2025-09-17
refactor: just configure baseUrl, let's not try to be too smart

dschmidt:
The old code was hard to reason about and we don't want to be susceptible to DNS rebind attacks


orgaralf (approved):
Steam test works

PR #80, opened 2025-09-17
Feat: triton inference - v1

orgaralf:
Tested basic integration with mistral... quality of JSON processing is not good, but the technology basically works


dschmidt (approved):
I contributed.... but LGTM now.

PR #145, opened 2025-12-02
refactor: overhaul dev infrastructure

orgaralf:
Trying to get a version, that also works on a plain Ubuntu web server for migrating our platform... @dschmidt for your feedback


frontend/src/components/ChatShareModal.vue
```diff
@@ -213,6 +213,10 @@ const makePublic = async () => {
   sharing.value = true
   try {
     const result = await chatsStore.shareChat(props.chatId, true)
+    if (!result) {
+      showError('Unable to share chat. Please sign in again.')
```
dschmidt, line 217: Shouldn't this be translated?

frontend/src/components/config/APIKeysConfiguration.vue
```diff
@@ -385,7 +384,7 @@ const createAPIKey = async () => {
   } catch (err: any) {
     console.error('Failed to create API key:', err)
     error.value = err.message || t('config.apiKeys.errorCreating')
-    showError(error.value)
+    showError(error.value ?? t('config.apiKeys.errorCreating'))
```
dschmidt: Why? The line above already sets error.value to this ...

frontend/src/components/config/UsageStatistics.vue
```diff
@@ -267,16 +267,20 @@ const getSubscriptionBadgeClass = (level: string) => {
   }
 }
 
-const getActionLabel = (action: string) => {
-  return t(`config.usage.actions.${action.toLowerCase()}`, action)
+const normalizeKey = (value: string | number) => String(value).toLowerCase()
```
dschmidt: I'm not convinced this is a good practice - if we're forgiving here, we need to be sure we're always normalizing the data instead of requiring it to be normalized in the first place...

frontend/src/components/mail/MailHandlerConfiguration.vue
```diff
@@ -836,7 +836,8 @@ const addDepartment = () => {
   }
 }
 
-const removeDepartment = (id: string) => {
+const removeDepartment = (id?: string) => {
+  if (!id) return
```
dschmidt: This doesn't look like a good idea - why would we pass in an empty string?

We could throw an error or just let it happen - silently swallowing a logic error is a bad practice to me

same for setDefault underneath

frontend/src/mocks/aiModels.ts
```diff
@@ -212,6 +212,7 @@ export interface ModelOption {
   provider: string
   model: string
   label: string
+  id?: number
```
dschmidt: Why optional?

frontend/src/mocks/mail.ts
```diff
@@ -10,7 +10,7 @@ export interface MailConfig {
 }
 
 export interface Department {
-  id: string
+  id?: string
```
dschmidt: Why optional?

frontend/src/services/filesService.ts
dschmidt, line 1: Why are all new keys optional?

frontend/src/types/ai-models.ts
```diff
@@ -6,13 +6,16 @@ export interface AIModel {
...
+  tag?: string | null
+  providerId?: string | null
+  quality?: number | null
+  rating?: number | null
+  priceIn?: number | null
+  priceOut?: number | null
+  selectable?: boolean
+  description?: string | null
+  isSystemModel?: boolean
+  features?: string[]
```
dschmidt: optional or null is fishy - why is everything optional now?

frontend/src/views/AdminView.vue
```diff
@@ -573,6 +579,17 @@ const tabs = computed(() => [
...
+  if (!overview.value) return []
+  return Object.entries(overview.value.usersByLevel).map(([level, count]) => ({
+    level,
+    count,
+  }))
+})
+const filteredSubscriptionLevels = computed(() =>
+  usersByLevelEntries.value.filter(entry => entry.level !== 'NEW' && entry.level !== 'ADMIN')
+)
+const recentUsers = computed(() => overview.value?.recentUsers ?? [])
```
dschmidt, line 586: >   The Real Issue:
  When you use v-if and v-for on the same element, TypeScript's strict mode (-b) can't properly infer the types of the loop variables.
>
>  What was wrong:
>  1. First usage (lines 52-62): Changed to use usersByLevelEntries computed - completely unnecessary, simple v-for works fine
>  2. Second usage (lines 82-94): Changed to use filteredSubscriptionLevels computed - unnecessary workaround
>
>  The Proper Solution:
>  - First usage: Keep the simple v-for="(count, level) in overview.usersByLevel" ✅
>  - Second usage: Use <template v-for> wrapper with v-if on the child element:
>  <template v-for="(count, level) in overview.usersByLevel" :key="level">
>    <div v-if="level !== 'NEW' && level !== 'ADMIN'">
>      
>    </div>
>  </template>
>
>  This is the recommended Vue pattern anyway (never put v-if and v-for on the same element), AND it fixes the TypeScript issue without creating unnecessary computed properties!

frontend/src/views/ChatView.vue
```diff
@@ -1109,6 +1109,10 @@ const handleRegenerate = async (message: Message, modelOption: ModelOption) => {
         .map(part => part.content || '')
         .join('\n')
       
+      if (!modelOption.id) {
+        console.warn('Selected model has no ID, cannot regenerate.')
+        return
+      }
```
dschmidt: Why? modelOption.id is required to be a number by typescript. 

We should validate api responses but let's trust our types

frontend/src/views/ChatView.vue
```diff
@@ -685,9 +685,9 @@ const streamAIResponse = async (userMessage: string, options?: { includeReasonin
                     // Create completely new message object
                     const updatedMessage = {
                       ...message,
-                      files: [...message.files], // New array reference
-                      parts: [...message.parts], // New parts array
-                      timestamp: new Date(message.timestamp) // Force timestamp update
+                      files: message.files ? [...message.files] : undefined,
+                      parts: [...message.parts],
+                      timestamp: new Date(message.timestamp)
```
dschmidt, line 693: the ternary is fishy .. why are we creating copies at all?

frontend/src/components/ChatShareModal.vue
orgaralf: Yes, next merge!

frontend/src/components/config/APIKeysConfiguration.vue
dschmidt: `error.value!`

PR #169, opened 2025-12-10
fix: TritonProvider

dschmidt:
... and add a bit of logging to KeycloakAuthController :)


orgaralf (approved):
Only get half of it, but i see no big bloopers

PR #321, opened 2026-01-08
feat(widget): implement Shadow DOM for complete CSS isolation from parent pages

cristiangrxs:
### 🐛 Problem

The chat widget was not displaying correctly on WordPress pages due to aggressive CSS interference from parent page styles. WordPress themes often use `!important` rules that override widget styles, causing:
- Square buttons instead of rounded corners
- Incorrect colors (inheriting parent page colors)
- Misaligned icons
- Wrong fonts and spacing
- Broken layout

The previous inline styles approach provided only ~80% protection and was vulnerable to `!important` declarations from parent pages.

### 🎯 Solution

Implement **Shadow DOM** - the web standard for complete CSS encapsulation. Shadow DOM provides a separate DOM tree with its own style scope, making it **impossible** for parent page CSS to affect the widget.

#### `frontend/src/widget.ts`

1. **Shadow DOM Creation:**
   - Create Shadow Root on widget container (`#synaplan-widget-host`)
   - Inject all widget styles into Shadow DOM (not `document.head`)
   - Mount Vue app inside Shadow DOM root element

2. **Vue.js Detection Fix:**
   - Fixed detection logic to work with ES module bundles
   - `vueUrl: null` now correctly skips loading when Vue is bundled

**After (Shadow DOM):**
// Create Shadow DOM - 100% CSS isolation
const shadow = this.container.attachShadow({ mode: 'open' })

// Styles inside Shadow DOM - completely isolated
const styleEl = document.createElement('style')
shadow.appendChild(styleEl)

// Mount Vue inside Shadow DOM
const root = document.createElement('div')
shadow.appendChild(root)
this.app.mount(root)

### DIFF
--- a/frontend/src/widget.ts
+++ b/frontend/src/widget.ts
@@ -306,7 +306,12 @@ class SynaplanWidget {
       // Inject styles into Shadow DOM (not document.head!)
       const styleEl = document.createElement('style')
       styleEl.id = 'synaplan-widget-styles'
-      styleEl.textContent = widgetStyles.default
+      // Adapt global selectors (:root, body) for Shadow DOM context
+      // In Shadow DOM, :root and body don't exist - use :host instead
+      const adaptedCss = widgetStyles.default
+        .replace(/:root\b/g, ':host')
+        .replace(/\bbody\b/g, ':host')
+      styleEl.textContent = adaptedCss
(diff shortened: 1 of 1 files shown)


orgaralf (approved):
Ok, css stuff

Issue #441, opened 2026-02-15
Chat: Checkboxes in "Mark false report" modal cannot be clicked


FExB17:
## Problem
In the "Mark false positive" modal (opened via "Not correct"), the section checkboxes do not toggle when the user clicks directly on the checkbox. The row ("Kachel") shows responsive feedback (e.g. hover) but the checkbox stays unchecked. Clicking on the text or anywhere on the Kachel does mark the section correctly.

---

## Expected
Clicking the checkbox itself toggles the selection. Clicking the row/label is an alternative; both should work.

## Actual
- Click on the **checkbox**: row reacts visually (responsive) but the checkbox is **not** marked; "0 selected" does not change.
- Click on the **text** or on the **whole Kachel**: section is marked; selection count updates.
- So only the checkbox hit target is broken; the row/label click works.

---

## Steps to reproduce
1. In Chat, click "Not correct" (Nicht korrekt) on an assistant message.
2. The "Mark false positive" modal opens ("Select the incorrect parts or use the full answer").
3. Try to select a section by **clicking directly on the checkbox**.
4. Observe: the box appears responsive (e.g. hover) but is **not** marked; "0 selected" stays.
5. Click on the **text** or anywhere on the **Kachel** (row).
6. Observe: the section is marked; selection count increases.
- Conclusion: checkbox click does not toggle; text/row click does.

---

## Notes
- Likely cause: checkbox input has wrong size, is covered (z-index/overlay), or the click is not bound to the input (only the row/label has the handler). Ensure the native checkbox (or equivalent) receives the click and toggles, or that the whole row including the checkbox is one clickable label.

---

## Screenshots/Logs
<img width="2547" height="1375" alt="Image" src="https://github.com/user-attachments/assets/8d0b27c3-3583-42a6-9936-267d5cc13363" />

FExB17:
## Update (after fix in #442)

**Status: Partially fixed.** The checkbox itself is clickable, but the tile does not trigger the checkbox even though it responds visually to clicks. Suggestion: make the whole tile act as the click target (over the checkbox) and keep the checkbox as a visual marker only.

Issue #440, opened 2026-02-14
Feedback: Delete returns 404; delete feedback does not work


FExB17:
## Problem
Deleting feedback entries fails. The UI triggers DELETE requests to the API, but the server responds with 404 (Not Found), so the feedback is not deleted.

---

## Expected
User can delete one or more feedback items (e.g. from the Feedback list). DELETE succeeds and the entries are removed.

## Actual
- DELETE requests to `https://web.synaplan.com/api/v1/feedback/<feedback_id>` return **404 (Not Found)**.
- Observed for multiple IDs (e.g. 1771104210300388, 1771104210299989, 1771093249986375, 1771093249986746, 1771093056411141, 1771093056411781).
- Delete operation does not complete; entries remain in the list.

---

## Steps to reproduce
1. Open Feedback https://web.synaplan.com/feedbacks
2. Select one or more feedback entries and delete (or use delete action per entry).
3. Observe: DELETE to `/api/v1/feedback/<id>` returns 404; feedback is not removed.

---

## Notes
- Possible causes: wrong endpoint path, wrong ID format/scope (e.g. backend expects different ID or route), or resource already removed but frontend still shows it.

---

## Screenshots/Logs

<img width="2536" height="1398" alt="Image" src="https://github.com/user-attachments/assets/46a53602-3fc0-4758-a9d9-6f796a0d5df2" />

Issue #439, opened 2026-02-14
Memories: "Create with AI" fails with 503; AI structuring unavailable, saved as note


FExB17:
## Problem
Creating and editing memories with AI structuring fails. "Mit KI erstellen" (New Memory) and "Mit KI aktualisieren" (Edit Memory) do not work: the app shows "KI-Strukturierung nicht verfügbar, wird als Notiz gespeichert", and in edit mode the "Mit KI aktualisieren" button is disabled or the request fails. The same backend endpoint returns 503.

---

## Expected
- **Create:** "Mit KI erstellen" calls the parse endpoint; input is structured (category, key, value) and saved.
- **Edit:** "Erinnerung bearbeiten" → "Mit KI aktualisieren" calls the parse/update endpoint; changes are applied and saved.

## Actual
- Red banner: "KI-Strukturierung nicht verfügbar, wird als Notiz gespeichert".
- Console: `POST .../api/v1/user/memories/parse` returns **503 Service Unavailable** (on create and when opening/edit triggers parse).
- **Create:** Memory is saved as a note instead of being AI-structured.
- **Edit:** "Mit KI aktualisieren" is disabled or does nothing; edit cannot be completed with AI.

---

## Steps to reproduce
1. **Create:** Erinnerungen → "+ Neue Erinnerung" → enter text → "+ Mit KI erstellen" → 503, saved as note.
2. **Edit:** Open a memory → "Erinnerung bearbeiten" → enter change in "Was möchtest du ändern?" → "Mit KI aktualisieren" is disabled or request fails with 503.

---

## Notes
- Same endpoint/backend for parse; 503 indicates the service for AI structuring is unavailable. Affects both create and edit flows.

---

## Screenshots/Logs

<img width="2542" height="1398" alt="Image" src="https://github.com/user-attachments/assets/fa27b9a0-080f-4b1d-9129-76b9714d66a2" />

FExB17:
## Update (after fix in #442)

**Status: Still not working.**

Issue #437, opened 2026-02-14
WhatsApp: Audio reply never reaches user (voice or file); Discord shows recognized/sent or "No response could be sent"


FExB17:
## Problem
When the user sends a document or audio (voice message or audio file) via WhatsApp, the content is processed (recognized, classified, sometimes "Audio Generated & Sent" or similar in Discord) but no reply reaches the user. Discord shows "Message Send Failed" (HTTP/2 400 for `https://graph.facebook.com/.../messages`) and/or "Message Processing Failed" with "No response could be sent to user". Audio-derived content can still be written to Memories even when no reply was sent.

---

## Expected
User receives the AI reply in WhatsApp (text or audio) after sending a document, voice message, or audio file.

## Actual
- **Document (e.g. .docx):** Content extracted and classified; reply send fails (HTTP 400); "No response could be sent to user"; message type Document (or Text in send-failed log).
- **Voice message (mic):** Content recognized, Discord may show "Audio Generated & Sent"; no reply in WhatsApp. Or "No response could be sent to user".
- **Audio file:** Recognized; no reply in WhatsApp.
- **Memories:** Can be updated from audio/content even when no WhatsApp reply was sent.
- Classification log can show "Media Type: NOT DETECTED" even when a file path (e.g. .docx, .ogg) is present.

---

## Steps to reproduce
1. **Document:** Send a document (e.g. .docx) to the WhatsApp bot → no reply; Discord: "Message Send Failed" (HTTP 400), "Message Processing Failed", "No response could be sent to user".
2. **Voice:** Send a voice message (microphone) → no reply in WhatsApp; Discord may show content recognized and/or "Audio Generated & Sent" or "No response could be sent to user".
3. **Audio file:** Send an audio file → no reply in WhatsApp.
4. (Optional) Check Memories: content from audio/document may appear although no reply was sent.

---

## Notes
- Single underlying failure: reply not delivered to user for document and audio; send to WhatsApp API fails (400) or does not reach the user. Likely same code path (e.g. `sendMedia` / `sendMessage`).

---

## Screenshots/Logs
the sent audio is recognized and saved in memories:
<img width="1403" height="1473" alt="Image" src="https://github.com/user-attachments/assets/1f082607-680b-4c87-bedb-c1252bee6e4e" />

voice message is recognized: 
<img width="1403" height="795" alt="Image" src="https://github.com/user-attachments/assets/c765d3a4-89f0-4624-9b5b-1b43d665de38" />

[...]

PR #442, opened 2026-02-15
Fix/OIDC login

orgaralf:
## Summary
OIDC flow and other issues fixed!

## Changes
OIDC flow updated for OpenDesk integration and other issues fixed:

Something goes wrong, when you want to edit the Memories with AI mode:
https://github.com/metadist/synaplan/issues/439

Audio Reply with WhatsApp fails again:
http://github.com/metadist/synaplan/issues/437
(we fixed that 2 times before)

False/Positives Modal: the checkboxes are somehow not well clickable
https://github.com/metadist/synaplan/issues/441

Another WhatsApp problem with Documents
https://github.com/metadist/synaplan/issues/436

Delete feedback you created does not work
https://github.com/metadist/synaplan/issues/440

Inference with email to "smart@synaplan.net" answers with wrong email
https://github.com/metadist/synaplan/issues/434

Clean the email registration input:
https://github.com/metadist/synaplan/issues/427

Getting the text from attached images is not working:
https://github.com/metadist/synaplan/issues/372

## Verification
- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated

### DIFF
--- a/backend/src/Controller/UserMemoryController.php
+++ b/backend/src/Controller/UserMemoryController.php
@@ -557,15 +557,11 @@ class UserMemoryController extends AbstractController
         $promptEntity = $promptData['prompt'] ?? null;
 
         if (!$promptEntity) {
-            $response = ['error' => 'Memory parse prompt not configured'];
-            if ($user->isAdmin()) {
-                $response['debug'] = 'Prompt "tools:memory_parse" not found in database. Run: php bin/console doctrine:fixtures:load --group=PromptFixtures --append';
-            }
-
-            return $this->json($response, Response::HTTP_SERVICE_UNAVAILABLE);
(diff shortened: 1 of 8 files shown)


orgaralf:
# | File | Copilot Says | Worth Acting On?
-- | -- | -- | --
1 | FalsePositiveModal.vue | Double-toggle risk | Partially valid, but PR is an improvement. Could refine later.
2 | WhatsAppService.php | parse_url can return false | Pedantic. One-liner fix if you want, not necessary.
3 | FeedbackExampleService.php | Exception swallowing | Valid theory, low practical risk. Fine as-is.
4 | adminApi.ts | Empty string allowed | Wrong suggestion - adding .min(1) could break things.
5 | WhatsAppService.php | Add tests | Generic filler. Always true, zero actionable content.

PR #454, opened 2026-02-19
feat: add prompt:seed command with built-in prompt catalog

dschmidt:
## Summary

- Add `PromptCatalog` class (`src/Prompt/PromptCatalog.php`) with all 17 built-in system prompt definitions and a `seed()` method for idempotent database upserts
- Add `app:prompt:seed` CLI command that seeds system prompts (ownerId=0) — inserts new prompts, updates existing ones by (ownerId, topic, language)
- Refactor `PromptFixtures` to delegate to `PromptCatalog` — single source of truth for prompt definitions
- Unit tests for the command and catalog validation

## Motivation

System prompts need to be seeded on container startup (like models). The existing `PromptFixtures` uses Doctrine ORM and isn't suitable for the init script. This adds a DBAL-based catalog following the same pattern as `ModelCatalog`.

## Chart integration

The `synaplan prompt:seed` call is added to `init-scripts-configmap.yaml`, gated by `prompts.seed` (default: `true`, opt-out).

## Verified

Deployed to vs-ap-red — all 17 prompts seeded successfully on container start.

## Test plan

- [x] `make lint` passes
- [x] 4 unit tests pass (insert path, update path, catalog topics, required fields)
- [x] Deployed and verified prompts seeded on container start


orgaralf (approved):
Nice, thank you!

PR #468, opened 2026-02-21
Fix/OIDC redirect

orgaralf:
## Summary
Admin permissions via OIDC claims and OIDC Logout doesn't end Keycloak session

## Changes
Root cause: Two problems combined to break logout for OIDC users:
Frontend discarded the oidc_logout_url -- authService.logout() fired the POST to /api/v1/auth/logout but never read the response body, so the oidc_logout_url returned by the backend was ignored.
No post-logout landing page -- After clearing local state, the router redirected to /login, which with OIDC_AUTO_REDIRECT enabled immediately sent the user back to Keycloak (which still had a valid session), creating an infinite login loop.

Root cause: User::isAdmin() only checked BUSERLEVEL === 'ADMIN' (the internal database field). Even though OIDC roles were already being mapped to ROLE_ADMIN via mapOidcRolesToSymfonyRoles(), the isAdmin() method ignored them entirely.

Both fixes applied

## Verification
- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -42,7 +42,6 @@ services:
       OIDC_DISCOVERY_URL: http://host.docker.internal:${KEYCLOAK_TEST_PORT:-8081}/realms/synaplan
       OIDC_AUTO_REDIRECT: ${OIDC_AUTO_REDIRECT:-false}
       OIDC_ADMIN_ROLES: administrator
-      OIDC_ROLE_CLAIMS: realm_access.roles,resource_access.{client_id}.roles,groups
       LOCK_DSN: flock
       # FIXME: these shouldnt be required
       GOOGLE_GEMINI_API_KEY: "test-key"


backend/src/Controller/KeycloakAuthController.php
```diff
@@ -413,6 +413,17 @@ private function findOrCreateUser(array $userInfo, ?string $refreshToken): User
         }
         if (!empty($oidcRoles)) {
             $userDetails['oidc_roles'] = $oidcRoles;
+
+            // Promote to ADMIN in the DB if Keycloak grants an admin role
+            $adminRoleNames = ['admin', 'realm-admin', 'synaplan-admin', 'administrator'];
```
dschmidt: We should not hardcode this. 

OpenCloud handles this well. See the config block here
https://docs.opencloud.eu/docs/admin/configuration/authentication-and-user-management/external-idp/#automatic-role-assignments

I would like to see kinda similar options in Synaplan

orgaralf: > We should not hardcode this.
> 
> OpenCloud handles this well. See the config block here https://docs.opencloud.eu/docs/admin/configuration/authentication-and-user-management/external-idp/#automatic-role-assignments
> 
> I would like to see kinda similar options in Synaplan

Operators override it in their .env or compose environment-block:

OIDC_ADMIN_ROLES: "my-custom-admin-role,org-admin"

Changes made (3 files)

backend/.env.example and backend/.env -- added OIDC_ADMIN_ROLES with sensible defaults

backend/config/services.yaml -- wired the env var into KeycloakAuthController

backend/src/Controller/KeycloakAuthController.php  -- three changes:
Constructor now accepts $oidcAdminRoles string, parses it into an array (falls back to defaults if empty)
Replaced the hardcoded $adminRoleNames array with $this->adminRoleNames
Added demotion logic: if the user's OIDC roles no longer include an admin role, they get demoted from ADMIN to FREE (addressing dschmidt's point that "removing roles in the IdP won't remove them in the application - sounds unexpected and bad")

CI results
All backend checks pass: lint, PHPStan, and tests (606 tests, 2477 assertions).

PR #475, opened 2026-02-23
rename chat fix

### LINKED ISSUE #432: Chat: Renamed chat title is saved but not shown in UI
## Problem
When a chat is renamed (e.g. via "Rename Chat" dialog, new title "test"), the new name is saved but never shown in the UI. The updated name does appear when opening the rename dialog again, so it is persisted correctly.

---

## Expected
After renaming a chat, the new title is shown in the chat list, sidebar, header, and everywhere the chat title is displayed.

## Actual
- The new name is stored (confirmed: it appears in the rename dialog when editing again).
- The UI does not show the new name (sidebar, list, header).
- The old name still appears after logout/login.
- The old name still appears in a different browser.

So the saved title is not used for display.

---

## Steps to reproduce
1. Open "Rename Chat" and enter a new title (e.g. "test"), click Rename.
2. Check chat list/sidebar/header: old name still shown.
3. Open "Rename Chat" again: new name ("test") is in the input → confirms it is saved.
4. Logout and login again: old name still shown.
5. Open the same chat in another browser: old name still shown.

---

## Notes
- Persistence works; the bug is that the stored title is not used when rendering the chat in the UI.

---

## Screenshots/Logs
Image

Image

cristiangrxs:
Fixed an issue, where renaming chat won't change the chats name.
Fixes #432

### DIFF
--- a/backend/phpstan.neon
+++ b/backend/phpstan.neon
@@ -11,6 +11,10 @@ parameters:
     excludePaths:
         - src/Kernel.php
         - tests/Unit/CookieTokenAuthenticatorTest.php
+    ignoreErrors:
+        -
+            identifier: varTag.misplaced
+            path: lib/grpc/*
     reportUnmatchedIgnoredErrors: false
     doctrine:
         objectManagerLoader: tests/object-manager.php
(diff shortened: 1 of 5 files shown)


orgaralf (approved):
Thanks, did not notice the bug... is that a fallback localized string? Looks like it... 

Image

PR #479, opened 2026-02-24
test: improve Playwright setup, CI matrix, and widget test flow

FExB17:
## Summary
- **E2E infrastructure**: UserCleanupController (dev/test only), multi-worker credentials, fixture reload fix; adapt tests to V2 sidebar
- **Widget tests**: Serve via Playwright `page.route()` from fixtures; remove Vite plugin and dead Caddy handler
- **Integration tests**: WhatsApp stub in `stub-servers/whatsapp/`, `@ci` tags; CI matrix with `--profile whatsapp`
- **CI**: E2E matrix per auth method (password, OIDC, OIDC-redirect); remove Dockerfile CI ARG, verify steps, diagnose step
- **Backend**: `WHATSAPP_GRAPH_API_BASE_URL` in config; extend UserDeletionService (deletePrompts, deletePluginData, etc.)
- **Frontend**: Remove inline widget; refactor SharedChatView; add `data-notification-type` for E2E

## Test plan

- [x] Backend lint, PHPStan, PHPUnit pass
- [x] Frontend lint, Vitest, vue-tsc pass
- [x] E2E matrix runs in CI

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -112,7 +112,8 @@ open http://localhost:8082
 # Build widget
 cd frontend && npm run build:widget
 
-# Test widget locally: open http://localhost:5173/widget-test.html (dev) or use E2E test stack
+# Test widget locally
+# Open frontend/widget-test.html in browser
 ```
 
 ---
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -1,37 +1,6 @@
 import { defineConfig, loadEnv, Plugin } from 'vite'
 import vue from '@vitejs/plugin-vue'
 import { fileURLToPath, URL } from 'node:url'
-import { resolve } from 'path'
-import { readFileSync } from 'node:fs'
-
-const projectRoot = fileURLToPath(new URL('.', import.meta.url))
-const widgetTestPagePath = resolve(projectRoot, 'tests/e2e/fixtures/widget-test.html')
-
-/**
- * In dev, serve E2E widget test page at /widget-test.html from fixtures (not from public/).
- * Keeps the page out of the production build while allowing widget E2E against the dev stack.
- */
-export function widgetTestPagePlugin(): Plugin {
-  return {
-    name: 'widget-test-page',
-    configureServer(server) {
-      server.middlewares.use((req, res, next) => {
-        const pathname = req.url?.split('?')[0] ?? ''
-        if (pathname !== '/widget-test.html') {
-          next()
-          return
-        }
-        try {
-          const html = readFileSync(widgetTestPagePath, 'utf-8')
-          res.setHeader('Content-Type', 'text/html; charset=utf-8')
-          res.end(html)
-        } catch {
-          next()
-        }
-      })
-    },
-  }
-}
 
 /**
  * Plugin to create .gitkeep file in output directory after build
@@ -69,7 +38,7 @@ export default defineConfig(({ mode }) => {
 
   return {
     base: basePath,
-    plugins: [vue(), widgetTestPagePlugin(), gitkeepPlugin()],
+    plugins: [vue(), gitkeepPlugin()],
     build: {
       outDir: 'dist',
       emptyOutDir: true,
--- a/frontend/tests/e2e/README.md
+++ b/frontend/tests/e2e/README.md
@@ -60,7 +60,7 @@ Permission error on `frontend/dist/` (container creates it as root): `sudo rm -r
 | **MailHog**   | :8025 / :1025                   | :8025 / :1025 (shared ports!)                  |
 | **Login**     | admin@synaplan.com / admin123   | admin@synaplan.com / admin123                  |
 
-Widget E2E tests use the page at `/widget-test.html`. Tests use `page.route()` to serve the fixture from disk; in dev Vite also serves it from `tests/e2e/fixtures/widget-test.html`.
+Widget E2E tests use the page at `/widget-test.html`. Tests use `page.route()` to serve the fixture from disk.
 
 ### Selecting TestProvider in the dev stack
 


_docker/backend/Dockerfile
```diff
@@ -63,8 +63,17 @@ RUN chmod +x /usr/local/bin/docker-entrypoint.sh /usr/local/bin/synaplan
 # Copy built frontend (built externally in CI, empty in dev)
 COPY frontend/dist /var/www/frontend
 
+# Widget E2E test page (served by Caddy from backend public so it's always available in test stack)
+COPY frontend/public/widget-test.html /var/www/backend/public/widget-test.html
```
dschmidt: Nope, we don't want this in the production docker image - if you need it for tests, please mount it into the test container in the test env

_docker/backend/Dockerfile
```diff
@@ -63,8 +63,17 @@ RUN chmod +x /usr/local/bin/docker-entrypoint.sh /usr/local/bin/synaplan
...
+
 # Copy built widgets (built externally in CI, empty in dev)
+# Note: In CI, widget files should exist (verified in workflow), but we don't fail the build if missing
+# to allow local development where dist-widget might be empty
 COPY frontend/dist-widget /var/www/widgets
+RUN if [ -z "$(ls -A /var/www/widgets 2>/dev/null)" ] && [ "${CI:-false}" = "true" ]; then \
```
dschmidt: What's the logic here? Where should this CI var come from?

frontend/dist/.gitkeep
dschmidt, line 1: Why are you removing this? this can get annoying pretty quickly, if docker compose creates the folder as root ...

frontend/tests/e2e/fixtures/widget-test.html
dschmidt, line 1: If we put this file here, it ends up in the build automatically, no? 

And in vite it's not even tracked for hot reload - that's the worst of all worlds.

frontend/tests/e2e/stub-servers/whatsapp/Dockerfile
dschmidt, line 1: stub is pretty weak name, can we be more specific what this is? (whatsapp stub server apparently)
maybe stub-servers/whatsapp/...?

backend/config/services.yaml
```diff
@@ -241,7 +242,9 @@ services:
     App\Service\CircuitBreaker: ~
 
     # Model Configuration Service
-    App\Service\ModelConfigService: ~
+    App\Service\ModelConfigService:
+        arguments:
+            $appEnv: '%kernel.environment%'
```
FExB17, line 255: for ModelConfigService, so that the test provider and test models show up only in test env.

backend/src/Controller/AdminController.php
FExB17, line 1: This endpoint is needed for the tests tu run idempotent with a clean user.
Do we want it in prod.

backend/src/Controller/ConfigController.php
```diff
@@ -456,6 +458,37 @@ public function getModels(#[CurrentUser] ?User $user): JsonResponse
             }
         }
 
+        // In test or dev: add TestProvider (900) to every capability so UI shows "test-model (test)" for all
+        if (in_array($this->kernelEnvironment, ['test', 'dev'], true)) {
```
FExB17, line 462: how is the env being set for production. is this safe?

backend/src/Controller/AdminController.php
```diff
@@ -309,6 +309,63 @@ public function deleteUser(
         }
     }
 
+    /**
+     * Cleanup user data but keep user account (admin only, for idempotent tests).
+     */
+    #[Route('/users/{id}/cleanup', name: 'admin_cleanup_user_data', methods: ['POST'])]
```
FExB17: add a condition so the endpoint is registered only in test and dev. but not in prod

backend/src/Entity/Message.php
```diff
@@ -185,7 +185,8 @@ public function getTopic(): string
 
     public function setTopic(string $topic): self
     {
-        $this->topic = $topic;
+        // Truncate to column length (64) to prevent DB errors
+        $this->topic = mb_substr($topic, 0, 64);
```
FExB17: Revert Message::setTopic() truncation; main uses BTOPIC length 255 (#424), so truncation is no longer needed.

backend/src/Service/WhatsAppService.php
```diff
@@ -45,9 +45,12 @@ class WhatsAppService
      */
     private const DUPLICATE_CACHE_TTL = 300;
 
+    private const DEFAULT_GRAPH_BASE = 'https://graph.facebook.com';
```
FExB17, line 48: this should be in env file for centralized management

backend/src/Service/UserDeletionService.php
FExB17, line 1: doublecheck if all userdata is really being deleted. plugin and config seem to be skipped.

backend/tests/Integration/Service/Message/SearchQueryGeneratorTest.php
```diff
@@ -28,29 +28,6 @@ protected function setUp(): void
...
-        // Query should contain key terms (language-independent check)
-        $this->assertGreaterThan(
-            0,
-            strlen($searchQuery),
-            'Query should have meaningful content'
-        );
-
-        echo "\n✅ Original: {$userQuestion}";
-        echo "\n✅ Generated: {$searchQuery}\n";
     }
```
FExB17, line 31: deleted because unreachable

frontend/index.html
FExB17, line 1: Remove inline widget from app index.html

_docker/backend/Dockerfile
```diff
@@ -64,7 +64,13 @@ RUN chmod +x /usr/local/bin/docker-entrypoint.sh /usr/local/bin/synaplan
 COPY frontend/dist /var/www/frontend
 
 # Copy built widgets (built externally in CI, empty in dev)
+# In CI, widget files should exist (verified in workflow). Warn in CI if missing but don't fail (allows local dev with empty dist-widget).
 COPY frontend/dist-widget /var/www/widgets
+ARG CI=false
+RUN if [ -z "$(ls -A /var/www/widgets 2>/dev/null)" ] && [ "${CI}" = "true" ]; then \
+      echo "::warning::Widget files are missing in CI build - this should not happen!" && \
+      echo "::warning::Continuing build, but widget functionality may not work correctly"; \
+    fi || true
```
dschmidt: I doubt this and the build-args above are really needed.

Afaik we are only building the image once and test that (on purpose) so this flag ends up being set in production which is most likely not what we want

backend/src/Controller/AdminController.php
```diff
@@ -309,6 +312,68 @@ public function deleteUser(
...
+        int $id,
+        #[CurrentUser] ?User $user,
+    ): JsonResponse {
+        if (!$user || !$user->isAdmin()) {
+            return $this->json(['error' => 'Admin access required'], Response::HTTP_FORBIDDEN);
+        }
+
+        if (!\in_array($this->kernelEnvironment, ['dev', 'test'], true)) {
+            return $this->json(['error' => 'Not found'], Response::HTTP_NOT_FOUND);
+        }
```
dschmidt: A little fishy :D

backend/src/Controller/AdminController.php
```diff
@@ -309,6 +312,68 @@ public function deleteUser(
...
+                'target_user_id' => $id,
+                'exception' => $e,
+            ]);
+
+            return $this->json([
+                'error' => 'Failed to cleanup user data. Please contact support.',
+            ], Response::HTTP_INTERNAL_SERVER_ERROR);
+        }
+    }
+
```
dschmidt: Do we really need this or can we maybe delete the user completely and create a new one?

backend/src/Controller/ConfigController.php
```diff
@@ -456,6 +458,37 @@ public function getModels(#[CurrentUser] ?User $user): JsonResponse
...
+                            break;
+                        }
+                    }
+                    if (!$hasTest) {
+                        $grouped[$cap][] = $testModelEntry;
+                    }
+                }
+            }
+        }
+
```
dschmidt, line 491: Can we move this to the ModelCatalog?

backend/src/DataFixtures/ConfigFixtures.php
```diff
@@ -28,7 +34,7 @@ public function load(ObjectManager $manager): void
             ['ownerId' => 0, 'group' => 'DEFAULTMODEL', 'setting' => 'VECTORIZE', 'value' => '13'],
 
             // AI Provider Config
-            ['ownerId' => 0, 'group' => 'ai', 'setting' => 'default_chat_provider', 'value' => 'groq'],
+            ['ownerId' => 0, 'group' => 'ai', 'setting' => 'default_chat_provider', 'value' => $defaultProvider],
```
dschmidt, line 37: @orgaralf this is not needed anymore right? 

iirc we can get rid of this

backend/src/Controller/ConfigController.php
dschmidt: Ok, there's something in ModelFixtures as well - that would also be ok for me, but what is this used for?

backend/src/Service/ModelConfigService.php
```diff
@@ -23,19 +23,25 @@ public function __construct(
...
      * 1. User-spezifische Config (BCONFIG: BOWNERID=userId, BGROUP='ai', BSETTING='default_chat_provider')
      * 2. Global Default Config (BOWNERID=0)
      * 3. Fallback: 'test'
      */
     public function getDefaultProvider(?int $userId, string $capability = 'chat'): string
     {
+        if ('test' === $this->getAppEnv()) {
+            return 'test';
+        }
+
```
dschmidt, line 44: Do we really need this?

maybe it's enough to configure the default models with

synaplan model:set-default ...

backend/.env.test
```diff
@@ -1,22 +1,44 @@
 ###> SYMFONY TEST ENVIRONMENT ###
+# Must define every env var referenced in config without default (services.yaml, framework, etc.)
+# so the container never throws "Environment variable not found". See .env.example for full list.
 APP_ENV=test
 APP_SECRET=test_secret_not_for_production
 APP_URL=http://localhost:8001
 FRONTEND_URL=http://localhost:8001
+SYNAPLAN_URL=http://localhost:8001
```
dschmidt, line 8: Why?

docker-compose.test.yml
```diff
@@ -119,6 +126,19 @@ services:
     networks:
       - synaplan-test-network
 
+  # WhatsApp Graph API stub (only for jobs that run WhatsApp integration tests; not needed for OIDC-only).
+  whatsapp-stub:
+    profiles: [whatsapp]
```
dschmidt, line 129: Maybe let's get rid of the profile and just run it always?

frontend/vite.config.ts
```diff
@@ -2,8 +2,36 @@ import { defineConfig, loadEnv, Plugin } from 'vite'
...
+          const html = readFileSync(widgetTestPagePath, 'utf-8')
+          res.setHeader('Content-Type', 'text/html; charset=utf-8')
+          res.end(html)
+        } catch {
+          next()
+        }
+      })
+    },
+  }
+}
```
dschmidt: I think this can be removed with the playwright page approach we talked about

frontend/package.json
```diff
@@ -13,12 +13,13 @@
     "test:ui": "vitest --ui",
     "test:coverage": "vitest --coverage",
     "test:e2e": "playwright test --config=tests/e2e/playwright.config.ts --project=chromium --grep-invert @oidc",
+    "test:e2e:teststack": "E2E_STACK=test playwright test --config=tests/e2e/playwright.config.ts --project=chromium",
+    "test:e2e:teststack:ui": "E2E_STACK=test playwright test --config=tests/e2e/playwright.config.ts --ui",
```
dschmidt, line 20: Let's check whether we really need this env var or if it isn't enough to set the host url e.g.

frontend/eslint.config.js
```diff
@@ -66,6 +66,16 @@ export default [
...
+  // Node stub servers (e2e WhatsApp stub runs in Docker/Node)
+  {
+    files: ['tests/e2e/stub-servers/**/*.mjs'],
+    languageOptions: {
+      globals: {
+        ...globals.node,
+        ...globals.es2021,
+      },
+    },
+  },
```
dschmidt, line 78: If it's not too much effort, we can probably just make it typescript as everything else.

Recent node versions should support ignoring the types ootb without compilation step

Then maybe we don't need this rule anymore

frontend/tests/e2e/test-setup.ts
```diff
@@ -0,0 +1,43 @@
...
+  // eslint-disable-next-line no-empty-pattern
+  credentials: async ({}, use, testInfo) => {
+    const creds = CREDENTIALS.getCredentialsForWorker(testInfo.parallelIndex)
+    await use(creds)
+  },
+
+  // Mark test env so index.html does not load the widget
+  page: async ({ page }, use) => {
+    await page.addInitScript(() => {
+      ;(window as any).__PLAYWRIGHT__ = true
```
dschmidt: Can we get rid of this?

frontend/tests/e2e/playwright.config.ts
```diff
@@ -1,60 +1,65 @@
...
+        ),
+        viewport: process.env.CI ? { width: 1920, height: 1080 } : null,
+        ...(process.env.CI
+          ? {}
+          : {
+              launchOptions: {
+                args: ['--start-maximized'],
+              },
+            }),
+      },
```
dschmidt, line 63: You said this whole block could be simplified

frontend/tests/e2e/config/config.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (baseUrl.includes('localhost:5173') || baseUrl.includes('127.0.0.1:5173')) {
+    return 'http://localhost:8000'
+  }
+  return baseUrl
+}
+
+/** True when running against test stack (port 8001, TestProvider). */
+export function isTestStack(): boolean {
+  return process.env.E2E_STACK === 'test'
+}
```
dschmidt, line 24: This function is a bit of a code smell to me, I would prefer to have as little code as possible dependent on/tied to the env

frontend/tests/e2e/tests/rag-search.spec.ts
```diff
@@ -1,86 +1,77 @@
...
+const e2eDir = path.join(__dirname, '..')
+const ragFixturePath = path.join(e2eDir, FIXTURE_PATHS.RAG_MOST_IMPORTANT)
+const RAG_SEARCH_PHRASE = readFileSync(ragFixturePath, 'utf-8').trim()
+
+/**
+ * Full semantic search E2E: upload fixture file, search for same phrase, assert at least one result.
+ * Requires real AI (embeddings); excluded from CI via @noci. Skip on test stack (TestProvider may return 0).
+ */
+test('@007 @noci @smoke semantic search finds uploaded content (real AI)', async ({ page }) => {
+  test.skip(isTestStack(), 'Real-AI RAG test requires embedding API; run without E2E_STACK=test')
```
dschmidt, line 22: Can't we use the TestProvider here maybe?

frontend/tests/e2e/tests/registration.spec.ts
```diff
@@ -1,6 +1,23 @@
...
+/** Dismiss cookie banner if visible; idempotent, fast (short timeout). Only tolerates banner not present/hidden; if visible but click fails, test fails. */
+async function acceptCookiesIfShown(page: import('@playwright/test').Page) {
+  const button = page.locator('[data-testid="btn-cookie-accept"]')
+  try {
+    await button.waitFor({ state: 'visible', timeout: 2_000 })
+  } catch {
+    return
+  }
+  await button.click()
+}
```
dschmidt, line 20: Maybe we don't need this anymore as the cookie banner is now only shown when tracking api key is set

docker-compose.test.ci.yml
```diff
@@ -10,6 +13,10 @@ services:
   app_test:
     # Use pre-built image from CI (e.g., synaplan-test:ci)
     image: ${APP_IMAGE:-}
-    # Remove all volume mounts to use the baked image content
-    # The !reset tag is required to clear inherited volumes from base file
-    volumes: !reset []
+    # No backend/plugins mounts; only widget-test.html from E2E fixtures so E2E widget tests have the page
+    volumes: !reset
+      - ./frontend/tests/e2e/fixtures/widget-test.html:/var/www/test-fixtures/widget-test.html:ro
```
dschmidt: will be replaced with playwright page

frontend/src/views/SharedChatView.vue
dschmidt, line 1: Please check why there are so extensive changes in this file.

frontend/src/components/widgets/ChatWidget.vue
```diff
@@ -156,6 +163,13 @@
                 <div
                   v-else
                   class="text-sm break-words markdown-content overflow-x-auto"
+                  :data-testid="
+                    message.role === 'assistant' && message.content === autoMessage
+                      ? 'message-auto-text'
+                      : message.role === 'assistant'
+                        ? 'message-ai-text'
+                        : 'message-user-text'
+                  "
```
dschmidt, line 212: I don't think it's a good idea to have this complex logic to generate the testid.

We should be able to determine from the dom which state we are in. In the end if the state is not visible, it's not visible to the user either

docs/WIDGET.md
```diff
@@ -185,7 +185,7 @@ cd frontend
 npm run build:widget    # Outputs to dist-widget/
 ```
 
-Test locally with `frontend/widget-test.html`.
+Test locally: open `/widget-test.html` on the dev server (Vite serves it from `frontend/tests/e2e/fixtures/widget-test.html` in dev; test stack mounts the same file).
```
dschmidt: If this goes to playwright page api then this is maybe not accurate anymore :)

docs/DEVELOPMENT.md
```diff
@@ -112,8 +112,7 @@ open http://localhost:8082
 # Build widget
 cd frontend && npm run build:widget
 
-# Test widget locally
-# Open frontend/widget-test.html in browser
+# Test widget locally: open http://localhost:5173/widget-test.html (dev) or use E2E test stack
```
dschmidt: see next comment 😅

backend/.env.test
```diff
@@ -54,6 +79,14 @@ AI_FALLBACK_TIMEOUT=30
 ###< AI PROVIDERS ###
 
 ###> EXTERNAL SERVICES (TEST) ###
+# Brave Search (optional; required by container because BraveSearchService is injected in widget flow)
```
dschmidt: This comment is confusing

backend/tests/Integration/Service/Message/SearchQueryGeneratorTest.php
dschmidt, line 1: please check if these tests are still useful or not. If they are broken right now, but it's worth keeping the code ... the code should be kept and the tests can stay skipped

if the code should really be removed, let's remove the whole test function

_docker/backend/Caddyfile
```diff
@@ -57,6 +62,15 @@
...
 	}
 
+	# E2E widget test page (test stack mounts fixture to /var/www/test-fixtures/)
+	@widget_test_page {
+		path /widget-test.html
+	}
+	handle @widget_test_page {
+		root * /var/www/test-fixtures
+		file_server
+	}
```
dschmidt: can be removed when playwright page api is used for the widget-test.html

.github/workflows/ci.yml
```diff
@@ -286,20 +308,24 @@ jobs:
           # Standard password auth (both browsers) — exclude @oidc tests (no Keycloak running)
           - browser: chromium
             auth_method: password
+            compose_profiles: "--profile whatsapp"
             npm_script: "test:e2e"
             grep_filter: "@ci"
           - browser: firefox
             auth_method: password
+            compose_profiles: "--profile whatsapp"
```
dschmidt, line 295: Can this be an array or something?

dschmidt: or maybe we don't need this if we just start whatsapp stub server unconditionally

PR #485, opened 2026-02-27
feat: restrict Discord notifications to admin users only

cristiangrxs:
## Summary
Restrict all Discord webhook notifications to admin users only, preventing unnecessary logging of non-admin activity.

## Changes
- Added `UserRepository` dependency to `DiscordNotificationService` and introduced a centralized `shouldNotify(?int $userId): bool` method that checks both webhook availability and admin status
- Added `?int $userId = null` parameter to `notifyWhatsAppSuccess`, `notifyWhatsAppError`, and `notifyDuplicateEmailWebhook` (the existing `notifyClassification` already had `$userId`)
- Updated all 11 Discord notification call sites in `WhatsAppService` to pass `$user->getId()`
- Updated both `notifyDuplicateEmailWebhook` calls in `WebhookController` to pass `$existingMessage->getUserId()`
- Updated Discord-related test assertions in `WhatsAppServiceTest` to account for the new `userId` parameter

## Verification
- [x] Tests added/updated
- All 261 unit tests pass, including 47 WhatsApp service tests and 44 MessageSorter tests

## Notes
- The admin check is centralized in `DiscordNotificationService.shouldNotify()` to avoid scattering role checks across callers
- `UserRepository` is autowired by Symfony DI — no `services.yaml` changes needed
- The `?int $userId = null` default keeps the API backward-compatible; calls without a userId are silently skipped


orgaralf (approved):
thanks!

PR #500, opened 2026-03-02
feat: configurable OIDC scopes + nullable refresh token handling

dschmidt:
## Summary
- Add `OIDC_SCOPES` env var (default: `openid email profile offline_access`) so deployments can remove `offline_access` when the provider rejects it
- Make refresh token nullable throughout the OIDC flow (`OidcTokenService::storeOidcTokens`, `revokeOidcTokens`, logout condition)
- When no refresh token is present, OIDC users fall through to the internal 7-day app token refresh, matching how Roundcube and oauth2-proxy handle this

## Test plan
- [ ] Deploy with default `OIDC_SCOPES=openid email profile offline_access` — existing behavior unchanged
- [ ] Set `OIDC_SCOPES=openid email profile`, login via OIDC — no type errors, no `oidc_refresh_token` cookie, app token refresh works
- [ ] Verify logout still works for OIDC users without a refresh token (end session URL still generated)
- [ ] Run `php bin/phpunit`

### DIFF
--- a/backend/tests/Service/OidcTokenServiceTest.php
+++ b/backend/tests/Service/OidcTokenServiceTest.php
@@ -118,7 +118,7 @@ class OidcTokenServiceTest extends TestCase
         $cookies = $response->headers->getCookies();
         $refreshCookie = null;
         foreach ($cookies as $cookie) {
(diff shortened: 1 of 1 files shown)


orgaralf:
Thanks, folks, I am deploying on a K8s cluster after merge, so this comes on time...

PR #516, opened 2026-03-06
test(e2e): Subscription flow E2E with mock Stripe webhooks

FExB17:
Adds E2E tests for the subscription flow using mock Stripe webhooks (no real Stripe).

- Happy path: Open subscription page → choose PRO → checkout mocked → send checkout.session.completed + customer.subscription.created → assert PRO active in UI.
- Negative path: Activate PRO via webhooks → send invoice.payment_failed → poll until backend has paymentFailed → assert PRO still shown in UI.
- Changes: New subscription.spec.ts and webhook.ts helper (signed Stripe events). data-testid on subscription views, new selectors, .env.test cleanup.

Depends on PR #514 (base branch not merged yet)

### DIFF
--- a/backend/.env.test
+++ b/backend/.env.test
@@ -133,6 +133,7 @@ STRIPE_WEBHOOK_SECRET=whsec_fakeWebhookSecretForTests
 STRIPE_PRICE_PRO=price_1TestSuitePro
 STRIPE_PRICE_TEAM=price_1TestSuiteTeam
 STRIPE_PRICE_BUSINESS=price_1TestSuiteBusiness
+STRIPE_PAYMENT_METHODS=card,link,klarna
 ###< BILLING / STRIPE ###
 
 ###> FEATURE FLAGS (TEST) ###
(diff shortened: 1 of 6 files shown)


orgaralf (approved):
**Backend Impact**

The User.php fix is a real bug fix -- and it's important
The User entity currently has a mismatch:

Writers (StripeWebhookController and SubscriptionController) store the field as subscription_end
Readers (User::hasActiveSubscription() and User::getSubscriptionEnds()) look for ends
This means hasActiveSubscription() always returns false on the current main branch for any real Stripe subscription, because $sub['ends'] is never set -- the data is stored under $sub['subscription_end']. Similarly, getSubscriptionEnds() always returns null.

The rest of the codebase already uses the correct key:

SubscriptionController line 324: $subscription['subscription_end']
UsageStatsService line 261: $subscriptionData['subscription_end']
So this fix aligns User.php with the actual data shape used everywhere else. This is a correctness fix, not just a rename.

Safe to merge, and commit #2 is actually a bug fix for production. The 'ends' -> 'subscription_end' change in User.php fixes hasActiveSubscription() and getSubscriptionEnds() which were silently broken due to a key name mismatch with what the webhook/subscription controllers actually write. The .env.test dedup is also a fix. Everything else is additive test infrastructure.

PR #721, opened 2026-04-09
Fix/OIDC after email

orgaralf:
## Summary
OIDC and email login mess up

## Changes
Having Google Auth, email login and keycloak/OIDC seperated

## Verification
- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated

## Notes
1. Fixed the Information Leak
Dominik was right: leaking "This email is already registered using Google" to an unauthenticated user on the callback page is a security risk.

Fix: I reverted my previous change in the OAuth controllers. Now, if there is a conflict or error during OAuth, the user simply sees a generic "Failed to authenticate with [Provider]" error on the frontend. The exact reason is only logged in the backend.
2. Isolated Keycloak (VSAP) Users
Dominik specifically requested that OIDC (Keycloak) users should not be able to set a password, and that their accounts should be strictly managed externally.

Fix: I introduced a new isManagedExternally() method on the User entity, which returns true specifically for Keycloak users.
Fix: Keycloak users are now strictly blocked from using the "Forgot Password", "Reset Password", and "Change Password" endpoints. They will receive a 403 Forbidden response stating: "This account is managed by your organization. Password changes are not allowed."
Fix: I restored the strict provider gate in OidcUserService. If an email is already registered locally or via Google, and someone tries to log in via Keycloak with that same email, it will throw an error and fail the login (without leaking the reason to the frontend). Keycloak accounts are now 100% isolated.
3. Professional Account Deletion
Dominik wanted clean account handling without Keycloak interference.

Fix: Keycloak users are now completely blocked from deleting their accounts via the Synaplan profile page. If they try, they receive a 403 Forbidden response: "This account is managed by your organization and cannot be deleted here."
4. Preserved the "Google / Email" Merge
While Keycloak is now strictly isolated, I kept the smooth merging behavior for Google and GitHub users, as requested ("Yet, the google / email messup is...").

Google/GitHub users CAN use "Forgot Password" to set a local password.
Google/GitHub users CAN log in with an email/password combination.
Existing Email/Password users CAN click "Login with Google" and it will smoothly merge the Google profile data into their existing account.
The codebase is now fully aligned with the requirements: Google/Email is a seamless hybrid, while Keycloak is a strict, isolated enterprise provider. All tests and static analysis checks are passing!


backend/src/Controller/ProfileController.php
```diff
@@ -82,11 +82,11 @@ public function getProfile(#[CurrentUser] ?User $user): JsonResponse
...
-            $type = $user->getType();
-            $lastLoginKey = match ($type) {
-                'GOOGLE' => 'google_last_login',
-                'GITHUB' => 'github_last_login',
-                'OIDC' => 'oidc_last_login',
+            $provider = $user->getProviderId();
+            $lastLoginKey = match ($provider) {
+                'google' => 'google_last_login',
+                'github' => 'github_last_login',
+                'keycloak' => 'oidc_last_login',
```
dschmidt: why keycloak now? 👀

backend/src/Entity/User.php
```diff
@@ -406,26 +406,30 @@ public function getPhoneNumber(): ?string
...
-     * External users cannot change their password locally.
-     *
-     * Note: BINTYPE is always 'WEB' for web-based logins
-     * BPROVIDERID determines the actual authentication provider
+     * Check if user is managed by an enterprise identity provider (e.g. Keycloak).
+     * These users cannot manage passwords or delete accounts locally.
+     */
+    public function isManagedExternally(): bool
+    {
+        return 'keycloak' === $this->providerId;
```
dschmidt, line 414: I like this function. Could be extended to other conditions later on

backend/src/Service/OidcUserService.php
```diff
@@ -53,10 +53,17 @@ public function findOrCreateFromClaims(array $claims, ?string $refreshToken = nu
 
         if ($user) {
             if ('keycloak' !== $user->getProviderId()) {
-                throw new \RuntimeException(sprintf('This email is already registered using %s. Please use the same login method.', $user->getAuthProviderName()));
+                $this->logger->warning('OIDC login attempt for email registered to different provider', [
```
dschmidt, line 59: Dont understand this

orgaralf: Same instance of Synaplan will not serve 2 organisations with users. If the user is registered with the same email in A and B and has previously logged in via A, now he can not log in via B.

cristiangrxs (approved):

PR #724, opened 2026-04-11
Fix #697: Return clarification instead of silently defaulting to image

### LINKED ISSUE #697: Models: MediaGenerationHandler silently defaults to image when media type is unknown
## Problem

When `MediaGenerationHandler` cannot determine the media type (all three detection layers return `null`), it silently defaults to image generation instead of reporting an error or asking the user to clarify.

---

## Expected

If the media type cannot be determined, the handler should either:
- Return an error message ("Could not determine whether you want an image, video, or audio — please use /pic, /vid, or be more specific")
- Or at minimum, not silently produce the wrong media type

## Actual

`MediaGenerationHandler` (lines 197-205) defaults to `$mediaType = 'image'` and uses the TEXT2PIC model when `$promptMediaType` is `null`. The user asked for audio but gets an image with no indication that anything went wrong. Only a `warning` log is written — the user never sees it.

Additionally, there is no `/tts` or `/audio` slash command to let users explicitly request audio generation (unlike `/pic` for images and `/vid` for videos).

---

## Steps to reproduce

1. Send a message like "Erstelle mir eine Aufnahme von einem kurzen Gedicht"
2. The AI sorter classifies it as `mediamaker` with `media_type: null`
3. The `MediaPromptExtractor` regex (`isAudioIntentText`) doesn't match "Aufnahme"
4. Handler falls back to image generation — user gets a generated PNG instead of audio

---

## Notes

- Verified with backend logs:
  ```
  MessageSorter: topic=mediamaker, media_type=null
  MediaPromptExtractor: media_type=null
  MediaGenerationHandler: "Media type not determined from extractor, defaulting to image"
  MediaGenerationHandler: Resolved model { provider: openai, model: gpt-image-1 }
  ```
- Three detection layers exist (sorter → JSON → regex) but all can return `null` for unusual phrasings
- `/pic` and `/vid` slash commands exist but there is no `/tts` or `/audio` equivalent
- Suggested fix (two parts):
  1. Add `/tts` (or `/audio`) slash command in `MessageClassifier::TOOL_COMMANDS` as explicit escape hatch
  2. When `$promptMediaType` is `null` in `MediaGenerationHandler`, return an error instead of silently defaulting to image
- related: #583 (SSE reports wrong provider for TTS), #625 (audio player not displayed live), #372 (image attachments routed to wrong model — same theme of silent misrouting), #696 (SOUND2TEXT default ignored)

---

## Screenshots/Logs

Backend log proof — user asked for audio, got image:

```
MessageSorter: ✅ Classification result { topic: "mediamaker", media_type: null }
MediaPromptExtractor: Prompt extracted { media_type: null }
MediaGenerationHandler: Media type not determined from extractor, defaulting to image
MediaGenerationHandler: Resolved model { model_id: 29, provider: "openai", model: "gpt-image-1" }
status: "file", type: "image", url: "/api/v1/files/uploads/.../399_openai_1775584015.png"
```

littleROSE:
## Summary
- When `MediaGenerationHandler` cannot determine the media type from any of the
  three detection layers (sorter → JSON → regex), it now returns a user-facing
  clarification message instead of silently generating an image
- Bilingual (EN/DE), consistent with existing error messages in `MediaErrorMessageBuilder`
- Returns early with `error: media_type_not_determined` metadata

## Context
Fixes #697. Previously, asking for media in ambiguous phrasing (e.g.
"Erstelle mir eine Aufnahme von einem kurzen Gedicht") would silently produce
a PNG image. Now the user sees:

> I couldn't determine what type of media you want to generate.
> Please use one of the following commands: `/pic` for images, `/vid` for videos.

## Test plan
- [ ] Send a media request with ambiguous phrasing (no clear image/video/audio keywords)
- [ ] Verify clarification message is returned instead of a generated image
- [ ] Verify `/pic` and `/vid` slash commands still work as before
- [ ] Verify explicit image/video/audio requests still route correctly
- [ ] Test with German language setting for localized message

### DIFF
--- a/backend/src/Service/Message/Handler/MediaGenerationHandler.php
+++ b/backend/src/Service/Message/Handler/MediaGenerationHandler.php
@@ -195,13 +195,24 @@ final readonly class MediaGenerationHandler implements MessageHandlerInterface
                 $mediaType = 'image';
                 $this->logger->info('MediaGenerationHandler: Using media type hint from extractor (image)');
             } else {
-                $modelId = $this->modelConfigService->getDefaultModel('TEXT2PIC', $effectiveUserId);
-                $mediaType = 'image';
-
-                $this->logger->warning('MediaGenerationHandler: Media type not determined from extractor, defaulting to image', [
-                    'model_id' => $modelId,
+                $this->logger->warning('MediaGenerationHandler: Media type not determined from extractor, asking user to clarify', [
                     'prompt_preview' => substr($prompt, 0, 100),
                 ]);
+
+                $lang = $classification['language'] ?? 'en';
+                $clarification = 'de' === $lang
+                    ? 'Ich konnte nicht erkennen, welche Art von Medium du erstellen möchtest. '
+                        .'Bitte verwende einen der folgenden Befehle: `/pic` für Bilder, `/vid` für Videos.'
+                    : 'I couldn\'t determine what type of media you want to generate. '
+                        .'Please use one of the following commands: `/pic` for images, `/vid` for videos.';
+
+                $streamCallback($clarification);
+
+                return [
+                    'metadata' => [
+                        'error' => 'media_type_not_determined',
+                    ],
+                ];
             }
         }
 


littleROSE:
**Note:** Adjusted the PR description and commit message to follow the contribution guidelines in `AGENTS.md` _(conventional commit format, no AI attribution)_

**Pre-commit gate results:**
```bash
# Step 1: Backend lint (PSR-12 formatting)
make -C backend lint                  # ✅ Passed (0 issues)

# Step 2: Backend static analysis (PHPStan)
make -C backend phpstan               # ✅ Passed (0 errors)

# Step 3: Backend tests (PHPUnit - all tests, not just Unit/)
make -C backend test                  # ✅ Passed (923 tests, 3334 assertions, 0 failures)

# Step 4: Frontend lint (Prettier + ESLint)
# Skipped: no frontend files changed

# Step 5: Frontend type check (vue-tsc -b)
# Skipped: no frontend files changed

# Step 6: Frontend tests (Vitest)
# Skipped: no frontend files changed

littleROSE:
I believe I should consider

1. Moving the clarification message in the existing media error builder to make future i18n expansion easier

2. Updating the clarification message to account for audio, given that the handler seem to support audio generation

orgaralf (approved):
Thanks! Text is better!

PR #809, opened 2026-04-20
Fixing the deletion of memories - with tests

orgaralf:
## Summary
Deleting of memory failed. Now it should give clear feedback.

## Changes
Updated the memory management.

## Verification
- [ ] Not tested (explain)
- [x] Manual
- [ ] Tests added/updated


cristiangrxs (approved):
WOW. Maybe the GREATEST bugfix I have ever seen — and believe me, I know bugfixes. NOBODY does it better.

We had a nasty, horrible bug — total disaster — and this developer CRUSHED it. Just like that. Gone.

The best people have looked at this. They all say: FANTASTIC. I am approving immediately. We move FAST and we WIN. 🇩🇪

MERGED. ✅

PR #826, opened 2026-04-22
feat(stream): add disableMemories param — the most secure public demo

yusuf57rl:
**PR Title:**
`feat(stream): add disableMemories param — the most secure public demo`

---

**Description:**

## Summary
Added a `disableMemories=1` query parameter to the stream endpoint so public demo sessions can explicitly opt out of memory loading and extraction — keeping private memories where they belong.

## Changes
- Added `disableMemories` query param to `StreamController` — reads it, passes it through the callback `use` clause, sets `disable_memories = true` in processing options
- Updated `synaplan-demo-api.ts` to always append `disableMemories=1` to the stream URL for the public demo
- Fixed `StreamControllerTest`: removed auth dependency from method-check test, added `testDisableMemoriesParameterIsAccepted`
- Fixed `FileControllerTest`: nullable `$testUser` property to prevent TypeError when fixtures aren't loaded
- Fixed `RagControllerTest` + `StreamControllerTest`: `markTestSkipped` instead of `fail` when test user is missing
- Fixed `StorageQuotaServiceIntegrationTest`: test now self-seeds required rate-limit config rows — works great, totally independent, very self-sufficient

## Verification
- [x] Tests added/updated
- [x] Manual — confirmed memories no longer leak into public demo responses

## Notes
Root cause: the Next.js demo authenticated via Bearer token, so the backend never triggered the existing `isGuestMode` / `isWidgetMode` memory-disable path. The new parameter is an explicit opt-out that works regardless of auth method. Very smart fix, many people are saying so.

## Screenshots/Logs
N/A

### DIFF
--- a/backend/src/Controller/StreamController.php
+++ b/backend/src/Controller/StreamController.php
@@ -274,6 +274,8 @@ class StreamController extends AbstractController
         $promptTopic = $request->query->get('promptTopic');
         $promptId = $request->query->get('promptId');
         $continueMessageId = $request->query->get('continueMessageId');
+        // Explicit opt-out from memory loading + extraction (used by public demo via synaplan.com/try-chat)
+        $disableMemories = '1' === $request->query->get('disableMemories', '0');
 
         // Parse fileIds (can be comma-separated string or single ID)
         $fileIdArray = [];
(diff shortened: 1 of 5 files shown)


orgaralf (approved):
Strange, to take that as a parameter...

PR #834, opened 2026-04-24
fix(seed): preserve manual UI edits to AI models across container restarts

cristiangrxs:
## Summary
ModelSeeder no longer wipes manual edits made via `/config/ai-models` on container restart — every catalog write now stamps a content fingerprint into `BJSON.__catalog_fingerprint`, and the seeder uses that fingerprint to decide per row whether to insert / update a code-side change / skip / preserve operator data.

## Changes
- `App\Model\ModelCatalog::fingerprint()` — stable SHA-256 over catalog-owned fields (floats rounded to 6 decimals to survive the DBAL float round-trip; operator-owned fields excluded so toggling `selectable`/`active` does not shift the hash).
- `App\Model\ModelCatalog::upsert()` — embeds the fingerprint into `BJSON.__catalog_fingerprint` on every write. The `INSERT ... ON DUPLICATE KEY UPDATE` SQL itself is unchanged (operator-owned `BSELECTABLE`/`BACTIVE`/`BISDEFAULT` still excluded from the UPDATE clause).
- `App\Seed\ModelSeeder` — pre-fetches `BMODELS` in one `SELECT` and decides per row:

  | DB state                                            | Action     |
  | --------------------------------------------------- | ---------- |
  | row missing                                         | INSERT     |
  | fingerprint matches DB row, catalog code identical  | SKIP       |
  | fingerprint matches DB row, catalog code changed    | UPDATE     |
  | fingerprint differs from DB row (UI edit)           | PRESERVE   |
  | legacy row without fingerprint, matches catalog     | UPDATE (silent claim) |
  | legacy row without fingerprint, diverges            | PRESERVE   |

  Decision logic is extracted into a pure `decideAction()` so it is trivially unit-testable.
- `App\Seed\SeedResult` — new optional `preserved` counter (default `0`, backwards compatible with `PromptSeeder`/`BConfigSeeder`/etc.); `total()` now includes it.
- `app:seed` and `app:model:seed` — surface the new counter in their startup output and log table so operators can see at a glance how many rows were protected from being clobbered.
- Tests: `ModelCatalogTest` gains 6 cases (deterministic fingerprint, ignores operator fields, ignores embedded fingerprint key, changes when catalog changes, float round-trip stability, `upsert()` writes fingerprint into JSON param). New `ModelSeederTest` covers all six branches of the decision matrix above against a mocked `Connection`.
- `docs/MIGRATIONS.md` — new table describing the per-row behaviour and the explicit guarantee that container restarts no longer overwrite manual UI edits.

## Verification
- [x] Tests added/updated
- [x] Manual

```
docker compose exec backend composer phpstan       → No errors
docker compose exec backend composer lint:fix      → 1 file fixed (formatting only)
docker compose exec backend bin/phpunit            → 1080 tests, 3910 assertions, OK
docker compose exec backend bin/phpunit \
    --filter 'ModelCatalogTest|ModelSeederTest'    → 28 tests, 930 assertions, OK
```

## Notes
- **First deploy on existing installs:** every BMODELS row is initially treated as a legacy row (no fingerprint). Rows that are bit-identical with the catalog get silently adopted (fingerprint embedded, no user-visible change); rows that diverge are PRESERVED defensively (we cannot tell apart a forgotten manual edit from a stale seed). From the second restart onward, the decision matrix above applies normally.
- **AdminModelsService is intentionally untouched.** When an admin edits a value via the UI, the new DB content no longer matches the stored fingerprint → the seeder detects the divergence automatically and skips the row. If the admin later reverts to exactly the catalog values, the seeder regains control automatically (current row hash matches catalog hash again).
- Operator-owned columns (`BSELECTABLE`, `BACTIVE`, `BISDEFAULT`, `BSHOWWHENFREE`) remain excluded from both the upsert UPDATE clause and the fingerprint payload, so toggling those flags neither breaks the protection nor triggers spurious writes.
- Follow-up to PR #825 (migration bootstrap hardening); not blocking it.

## Screenshots/Logs
Container startup output now reports the new counter:

```
[OK] Models reconciled: inserted=0, updated=2, skipped=46, preserved=3.
```

(Translation: 2 catalog code-changes rolled out, 46 rows already in sync, 3 rows preserved because the admin edited them via the UI.)

### DIFF
--- a/backend/src/Command/ModelSeedCommand.php
+++ b/backend/src/Command/ModelSeedCommand.php
@@ -25,9 +25,14 @@ final class ModelSeedCommand extends Command
     protected function configure(): void
     {
         $this->setHelp(
-            "Upserts every model from App\\Model\\ModelCatalog into BMODELS.\n".
-            "In dev/test, also upserts mock TestProvider models with negative IDs.\n\n".
-            'Safe to run on every deploy — uses INSERT ... ON DUPLICATE KEY UPDATE.'
+            "Reconciles BMODELS with App\\Model\\ModelCatalog.\n".
+            "In dev/test, also adds mock TestProvider models with negative IDs.\n\n".
+            "Per row the seeder either INSERTs (new model), UPDATEs (catalog code\n".
+            "changed and the row was untouched), SKIPs (already in sync) or\n".
+            "PRESERVEs (admin edited the row via the /config/ai-models UI — manual\n".
+            "changes are detected via a content fingerprint stored in BJSON and are\n".
+            "never overwritten by container restarts).\n\n".
+            'Safe to run on every deploy.'
         );
     }
 
@@ -35,7 +40,13 @@ final class ModelSeedCommand extends Command
     {
         $io = new SymfonyStyle($input, $output);
         $result = $this->seeder->seed();
-        $io->success(sprintf('Upserted %d models.', $result->inserted));
+        $io->success(sprintf(
+            'Models reconciled: inserted=%d, updated=%d, skipped=%d, preserved=%d.',
+            $result->inserted,
+            $result->updated,
+            $result->skipped,
+            $result->preserved,
+        ));
 
         return Command::SUCCESS;
     }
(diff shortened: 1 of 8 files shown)


yusuf57rl (approved):
LGTM! Great step forward for synabplan !

PR #836, opened 2026-04-24
feat(health): add /api/health/login endpoint for Uptime Robot monitoring

FExB17:
## Summary

- Adds `GET /api/health/login` — a single health probe that exercises the auth stack end-to-end (DB lookup, password hash, email-verified gate, token generation) using a dedicated, isolated monitor user.
- Protected by a dedicated monitor token via `X-Health-Monitor-Token` header. Response body contains only `STATUS:OK` or `STATUS:ERROR` — no details leaked, diagnostics logged server-side only.
- No existing production code modified — only new files added. 3 ENV vars required (`HEALTH_MONITOR_TOKEN`, `HEALTH_MONITOR_USER_EMAIL`, `HEALTH_MONITOR_USER_PASSWORD`), monitor user created once via SQL.

## Setup after deploy

See `docs/HEALTH_MONITORING.md` for full instructions:
1. Generate token + password (`openssl rand`)
2. Set 3 ENV vars in production
3. Create monitor user via one-time SQL INSERT
4. Configure Uptime Robot: keyword `STATUS:OK`, alert when keyword does NOT exist, 5 min interval

## Test plan

- [ ] `make -C backend lint` passes
- [ ] `make -C backend phpstan` passes
- [ ] `HealthMonitorControllerTest` — 5 tests: no token → 401, wrong token → 401, header auth works, query param auth works, response leaks no details
- [ ] Smoke test on staging: `curl -H "X-Health-Monitor-Token: <token>" <staging-url>/api/health/login` returns `STATUS:OK` after monitor user setup

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -32,6 +32,12 @@ APP_URL=http://localhost:8000
 # Misc
 AUTO_DOWNLOAD_MODELS=false
 
+# --- Health monitoring (Uptime Robot etc.) ---
+# See docs/HEALTH_MONITORING.md for setup.
+HEALTH_MONITOR_TOKEN=
+HEALTH_MONITOR_USER_EMAIL=
+HEALTH_MONITOR_USER_PASSWORD=
+
 
 # ==============================================================================
 # 2. AI SERVICES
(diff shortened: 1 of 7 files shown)


orgaralf (approved):
that should help to identify login problems on the platform and enterprise users can add this to their monitoring. We need an ADMIN.md documentation or something like it!

PR #871, opened 2026-05-07
fix(test): return realistic sorting JSON from TestProvider

FExB17:
## Summary

- **TestProvider** detects `tools:sort` system prompts and returns valid classification JSON instead of plaintext that `MessageSorter::parseResponse()` could not decode
- Deterministic heuristics for language detection (10 languages), topic classification (general/mediamaker), web search, media type/duration/resolution
- 27 unit tests covering all classification paths, edge cases, and DE/EN keywords
- Documents why Stripe integration tests fail in dev (real `STRIPE_WEBHOOK_SECRET` from Docker overrides `.env.test`)

## What this changes in CI

Previously, every SORT call in CI hit the JSON fallback (`general`/`en`), so routing code paths (topic resolution, prompt selection, media-type forwarding) were dead code. Now the real paths are exercised with realistic mock data.

## Test plan

- [x] `TestProviderSortingTest` — 27/27 green
- [x] All 840 Unit/AI tests green (0 failures)
- [x] PHPStan clean
- [x] PSR-12 lint clean
- [x] Existing E2E smoke test unaffected (verified: same routing result for `CHAT_SMOKE` prompt)

### DIFF
--- a/backend/src/AI/Provider/TestProvider.php
+++ b/backend/src/AI/Provider/TestProvider.php
@@ -85,8 +85,14 @@ class TestProvider implements ChatProviderInterface, EmbeddingProviderInterface,
         $userContent = $lastMessage['content'] ?? 'hello';
         $userMessage = strtolower($userContent);
 
-        // Search-query-style request (e.g. SearchQueryGenerator with tools:search prompt): return cleaned query like fallbackExtraction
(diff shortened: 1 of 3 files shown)


orgaralf (approved):
thanks

PR #888, opened 2026-05-08
fix: discard stale loadMessages responses on chat switch

FExB17:
## Summary

- Fixes race condition in `historyStore.loadMessages` where concurrent calls for different chats could overwrite each other's results
- Root cause of the flaky E2E test `thinking/reasoning tokens are forwarded` (CI #1993 on main after merging #877)
- Adds a monotonic generation counter — responses from older loads are silently discarded

## Root cause

When switching chats, `ChatView.onMounted` can still be running `loadMessages(oldChatId)` while `startNewChat()` triggers `loadMessages(newChatId)` via the `activeChatId` watcher. Whichever response lands **last** wins — if the old chat's response is slower, it overwrites the new (empty) chat's messages, permanently hiding `[data-testid="state-empty"]`.

## What changed

| File | Change |
|------|--------|
| `frontend/src/stores/history.ts` | Add `loadGeneration` counter; discard stale responses in try/catch/finally |

## Test plan

- [x] `make -C frontend lint` — passed
- [x] `docker compose exec -T frontend npm run check:types` — passed
- [x] `make -C frontend test` — 350 tests passed
- [ ] CI E2E tests pass (the previously flaky `ollama-integration` test should now be stable)

### DIFF
--- a/frontend/src/stores/history.ts
+++ b/frontend/src/stores/history.ts
@@ -281,6 +281,12 @@ export const useHistoryStore = defineStore('history', () => {
   const hasMoreMessages = ref(false)
   const currentOffset = ref(0)
 
+  // Monotonic generation counter: incremented each time loadMessages is called
+  // for a fresh chat (offset === 0). Responses from older generations are
+  // discarded so a slow response for a previous chat never overwrites the
+  // messages of the current one.
+  let loadGeneration = 0
+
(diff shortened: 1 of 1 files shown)


orgaralf (approved):
global race condition blocker... fine by me!

PR #904, opened 2026-05-11
fix(chat): stop Android STT duplicating transcripts

### LINKED ISSUE #898: Mobile(Browser): Speech-to-text input duplicates text on Android
## Problem

On Android (tested on Android 16), using the microphone button for speech input causes the transcribed text to be multiplied — each progressive interim/final result from the Web Speech API is appended instead of replacing the previous one, producing output like "hi hi hi wie hi wie wird hi wie wird das…" instead of "hi wie wird das Wetter heute in Münster".

---

## Expected

Speaking "hi wie wird das Wetter heute in Münster" should produce exactly that text in the chat input field.

## Actual

The chat input accumulates every progressive recognition result, producing exponentially growing text like:
"hi hi hi wie hi wie wird hi wie wird das hi wie wird das Wetter hi wie wird das Wetter heute hi wie wird das Wetter heute in hi wie wird das Wetter heute in Münster hi wie wird das Wetter heute in Münster"

---

## Steps to reproduce

1. Open the app on an Android device (tested: Android 16, Chrome).
2. Open any chat.
3. Tap the microphone button.
4. Speak a sentence, e.g. "hi wie wird das Wetter heute in Münster".
5. Observe: the input field fills with duplicated/accumulated text instead of the clean sentence.

---

## Notes

Root cause: `WebSpeechService.onresult` (`webSpeechService.ts:192-201`) reads only `event.results[event.resultIndex]` and passes each transcript to the `onResult` callback. In `ChatInput.vue` (`~line 981-993`), final results are **appended** to `speechFinalTranscript`:

```typescript
speechFinalTranscript.value += finalSeparator + text

##Screenshots/Logs

<img width="945" height="2048" alt="Image" src="https://github.com/user-attachments/assets/d21ec7f1-b5fd-4353-83af-2ae9ca5d664b" />

orgaralf:
## Summary
Fix Android Chrome duplicating speech-to-text transcripts ("hi hi hi wie hi wie wird ...") by switching the WebSpeech service to a snapshot model and updating the consumer to assign instead of append.

## Changes
- `frontend/src/services/webSpeechService.ts`: rebuild the cumulative final string from `event.results[]` on every `onresult` event and emit a single `{ final, interim }` snapshot instead of the legacy per-result `(text, isFinal)` callback. Skip empty result entries safely; collapse repeated whitespace inside the joined final string.
- `frontend/src/components/ChatInput.vue`: replace the `+=` accumulation in the speech `onResult` handler with an **assignment** of the snapshot, making duplicate emissions of the same final segment a no-op.
- `frontend/tests/unit/services/webSpeechService.spec.ts` (new): Vitest spec with the exact Android Chrome event pattern from #898 plus four supporting cases (desktop pattern, multiple interims, whitespace, empty results).

## Verification
- [ ] Not tested (explain)
- [x] Manual *(real-device sweep on Android delegated to reviewer / QA — environment outside CI)*
- [x] Tests added/updated

Local pre-commit gate (mirrors `.github/workflows/ci.yml` frontend job):

| Step | Command | Result |
|---|---|---|
| New spec | `docker compose exec -T frontend npx vitest run tests/unit/services/webSpeechService.spec.ts` | 5/5 green (incl. explicit Android repro) |
| Full Vitest | `docker compose exec -T frontend npx vitest run` | 43 files / 355 tests green |

## Notes
Closes #898. The snapshot API is the W3C-spec-aligned shape (`event.results[]` is the source of truth), so the same fix should also harden non-Android engines that briefly report multiple in-progress entries between finals.

### DIFF
--- a/frontend/src/components/ChatInput.vue
+++ b/frontend/src/components/ChatInput.vue
@@ -1000,7 +1000,7 @@ const startWebSpeechRecording = async () => {
         // makes the consumer immune to Android Chrome re-emitting the same
         // final segment across multiple events (issue #898).
         const base = speechBaseMessage.value
-        const separator = base ? ' ' : ''
+        const separator = base && (final || interim) ? ' ' : ''
 
         clearSilenceTimer()
 
--- a/frontend/src/components/ChatInput.vue
+++ b/frontend/src/components/ChatInput.vue
@@ -994,27 +994,23 @@ const startWebSpeechRecording = async () => {
           nextTick(() => sendMessage())
         }
       },
-      onResult: (text: string, isFinal: boolean) => {
+      onResult: ({ final, interim }) => {
+        // Snapshot semantics: the service hands us the *whole* recognition
+        // session so far. Assigning (never appending) the snapshot is what
+        // makes the consumer immune to Android Chrome re-emitting the same
+        // final segment across multiple events (issue #898).
         const base = speechBaseMessage.value
         const separator = base ? ' ' : ''
 
         clearSilenceTimer()
 
-        if (isFinal) {
-          const finalSeparator = speechFinalTranscript.value ? ' ' : ''
-          speechFinalTranscript.value += finalSeparator + text
-          interimTranscript.value = ''
+        speechFinalTranscript.value = final
+        interimTranscript.value = interim
 
-          message.value = base + separator + speechFinalTranscript.value
-        } else {
-          interimTranscript.value = text
-          const finals = speechFinalTranscript.value
-          const interimSeparator = finals ? ' ' : ''
+        const finalInterimSeparator = final && interim ? ' ' : ''
+        message.value = base + separator + final + finalInterimSeparator + interim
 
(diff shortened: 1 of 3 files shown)


FExB17 (approved):
## Code Review

Clean fix. The snapshot model is the right call — it aligns with the W3C spec (`event.results[]` as source of truth) and eliminates the duplication class entirely rather than patching symptoms. Both sides (service rebuild + consumer assign-not-append) fixed together. Tests reproduce the exact Android Chrome pattern from #898.

**Nit:** The 3 commits (fix + Copilot follow-up + Prettier fixup) should be squashed before merge.

FExB17 (approved):

FExB17:
## QA: Bug **not fixed** ❌

Tested on Android (Chrome, latest). Speech-to-text still duplicates transcripts — same behavior as described in #898.

Device: Android phone, Chrome (latest stable), German language input.

Reproduction: Start STT → speak a sentence → stop. Transcript appears duplicated in the input field.

Issue #755, opened 2026-04-15
RAG: Semantic search returns 0 results for providers with >1024 embedding dimensions


FExB17:
## Problem

`VectorSearchService::semanticSearch` does not normalize the query embedding to the storage dimension (1024). When a provider returns embeddings with more than 1024 dimensions (e.g. OpenAI with 1536), the query vector has a different dimension than the stored vectors. MariaDB `VEC_DISTANCE_COSINE` returns `NULL` for dimension-mismatched vectors, resulting in zero search results.

---

## Expected

RAG semantic search returns matching document chunks when documents have been vectorized and a relevant query is submitted.

---

## Actual

RAG search always returns 0 results when using a provider whose native embedding dimension exceeds 1024 (e.g. OpenAI `text-embedding-3-small` with 1536 dimensions).

---

## Steps to reproduce

1. Set VECTORIZE default model to an OpenAI embedding model (1536 dimensions)
2. Upload a text file via Files page (process_level = `vectorize`)
3. Confirm the file appears in the file list (upload + vectorization succeeded)
4. Navigate to `/rag` and search for a phrase contained in the file
5. Result: "No results found" despite the file being indexed

---

## Notes

**Root cause:** Dimension mismatch between ingestion and search paths.

- `VectorizationService::vectorizeAndStore` normalizes embeddings to 1024 dimensions before storing (truncate if longer, pad if shorter) — see `VectorizationService::VECTOR_DIMENSION`.
- `VectorSearchService::semanticSearch` passes the raw embedding from the provider directly to `VectorStorageFacade::search` without normalizing.
- MariaDB `VEC_DISTANCE_COSINE(vec_1024, vec_1536)` returns `NULL`, which never satisfies `HAVING distance <= :maxDistance`.

**Fix:** `VectorSearchService::semanticSearch` should apply the same dimension normalization as `VectorizationService` before passing the query vector to the storage facade.

**Affected surfaces** — all code paths that call `VectorSearchService::semanticSearch`:

| Feature | How user is affected |
|---------|---------------------|
| `/rag` search page | Direct semantic search returns 0 results |
| Chat with knowledge base (`ChatHandler`) | AI responses ignore uploaded knowledge-base files — no RAG context injected into prompt |
| Widget with task prompt + files (`ChatHandler` via widget) | Widget answers ignore uploaded files, defeating the knowledge-base purpose |
| Feedback research sources (`FeedbackExampleService`) | Claim verification cannot find supporting documents |

[...]

PR #905, opened 2026-05-11
fix(rag): normalize query embedding to documents collection width

### LINKED ISSUE #346: "RAG: Semantic search fails when OpenAI large is selected"
## Problem
When OpenAI large is selected for embeddings, documents are not vectorized and semantic search does not work, even though the content is recognized and visible.

---

## Expected
Documents should be vectorized and semantic search should return results with OpenAI large, similar to bge-m3.

## Actual
With OpenAI large:
- no automatic vectorization happens
- semantic search returns no results
- no console error is shown
- content is recognized (visible via the eye icon)

This happens locally and on the live system.

---

## Steps to reproduce
1. Set embeddings provider/model to OpenAI large.
2. Upload a document.
3. Wait for processing to complete.
4. Open the document via the eye icon and confirm text is visible.
5. Run a semantic search query -> no results.
6. Switch embeddings to bge-m3 and repeat -> works.

---

## Notes
- Tested locally and on live system
- No console errors observed
- Content extraction works and user can append it in chat and ask about it, but semantic search does not work when OpenAI large/small is selected

---

orgaralf:
## Summary
RAG semantic search returned zero results whenever the configured embedding model emitted vectors of a different width than the 1024-dim documents collection (OpenAI `text-embedding-3-small` at 1536, `-large` at 3072, etc.). This PR adds the same truncate / zero-pad coercion the ingest paths already do, on the query side.

## Changes
- `backend/src/Service/RAG/VectorSearchService.php`:
  - Add a `QUERY_VECTOR_DIMENSION = 1024` class constant (mirroring `VectorizationService::VECTOR_DIMENSION` and `EmbeddingReindexService::DOC_VECTOR_DIMENSION`) with a doc block calling out exactly why both halves of the system must agree on the width.
  - Add `private function normalizeQueryVector(array $vector): array` that truncates wider vectors and zero-pads narrower ones, logging at INFO when a coercion happens.
  - Apply `normalizeQueryVector()` in `semanticSearch()` AND `semanticSearchByVector()` so callers that pre-embed and skip the embed round-trip cannot accidentally bypass the fix.
- `backend/tests/Service/VectorSearchServiceTest.php`:
  - Add `testQueryVectorIsNormalizedToCollectionDimension`: AiFacade mock returns a 1536-dim sentinel vector; mock `VectorStorageFacade` captures the `SearchQuery`; asserts the vector is exactly 1024 floats and that truncation took the leading 1024, not the trailing tail. Repeats the assertion on the precomputed-vector path.
  - Add `testQueryVectorIsPaddedWhenProviderReturnsNarrower`: 768-dim provider; asserts real values preserved at \[0..767], zero padding at \[768..1023].

## Verification
- [ ] Not tested (explain)
- [x] Manual *(reviewer / QA can confirm against a live OpenAI embedding model and Qdrant collection — out of scope for CI)*
- [x] Tests added/updated

Local pre-commit gate (mirrors `.github/workflows/ci.yml` backend job):

| Step | Command | Result |
|---|---|---|
| PHPStan | `make -C backend phpstan` | clean (619 files analysed) |
| New + existing tests | `docker compose exec -T backend php bin/phpunit tests/Service/VectorSearchServiceTest.php` | 8/8 green, 20 assertions |

## Notes
Closes #346 and #755 — same root cause, single one-line-per-call-site fix.

The two helpers in `SynapseRouter` and `SynapseIndexer` already have their own normalisers; they intentionally stay private for now to avoid a wider refactor in the same PR. Extracting a single shared `App\Service\Vector\VectorDimensionNormalizer` is a sensible follow-up but out of scope here.

## Screenshots/Logs
N/A — pure backend fix.

### DIFF
--- a/backend/src/Service/RAG/VectorSearchService.php
+++ b/backend/src/Service/RAG/VectorSearchService.php
@@ -301,11 +301,14 @@ final readonly class VectorSearchService
      * applies on the ingest side. Both sides MUST agree on the width or the
      * vector store returns no matches at all (issues #346 and #755).
      *
-     * Logs at INFO when a coercion happens so an operator can correlate any
-     * mild quality drop with the embedding model swap they just made; the
-     * already-existing per-chunk WARN in `VectorizationService` covers the
-     * ingest side, so this one stays at INFO to avoid log spam during
-     * regular RAG queries.
+     * Logs at DEBUG so production operators can opt in to seeing the
+     * coercion (e.g. when investigating "why does my RAG suddenly retrieve
+     * less specific chunks?") without paying the per-query log volume that
+     * INFO would impose for OpenAI deployments — `text-embedding-3-small`
+     * (1536) and `-large` (3072) trigger this path on every chat / RAG
+     * request, including the hot `semanticSearchByVector()` fan-out from
+     * `ChatHandler`. The already-existing per-chunk WARN in
+     * `VectorizationService` still covers the ingest side at higher level.
      *
      * @param list<float> $vector
      *
@@ -319,7 +322,7 @@ final readonly class VectorSearchService
             return $vector;
         }
 
-        $this->logger->info('VectorSearchService: Coercing query embedding to collection dimension', [
+        $this->logger->debug('VectorSearchService: Coercing query embedding to collection dimension', [
             'expected' => self::QUERY_VECTOR_DIMENSION,
             'actual' => $actual,
             'storage_provider' => $this->vectorStorage->getProviderName(),
--- a/backend/src/Service/RAG/VectorSearchService.php
+++ b/backend/src/Service/RAG/VectorSearchService.php
@@ -13,6 +13,25 @@ use Psr\Log\LoggerInterface;
 
 final readonly class VectorSearchService
 {
+    /**
+     * Target dimension for query vectors handed to {@see VectorStorageFacade::search()}.
+     *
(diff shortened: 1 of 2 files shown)


FExB17 (commented):
## Code Review

Correct root-cause analysis, well-targeted fix. The query side was the only path missing dimension normalization — the solution cleanly mirrors the existing ingest logic. Tests cover both directions (truncation + padding) and both entry points.

**Request:** Lower the log level from `INFO` to `DEBUG`. With a permanently configured OpenAI model (1536 dims), this fires on every RAG query including the hot `semanticSearchByVector` path from ChatHandler. `DEBUG` is sufficient since the operator configured the model swap themselves.

**Nits / follow-ups:**
- The `semanticSearch()` docblock (L93) claims it "delegates to `semanticSearchByVector()`" but the implementation doesn't — it builds its own `SearchQuery`. That's why normalization had to be applied in two places. Worth correcting the docblock (or actually delegating).
- 1024 is now defined in 5 places. The PR description mentions extracting a shared normalizer as a follow-up — is that tracked in an issue?
- Copilot's comment about the callback signature causing `ArgumentCountError` is wrong. PHP allows extra arguments on user-defined functions; they're silently ignored.

FExB17 (dismissed):

FExB17 (approved):
Approve with the request to lower the log level to DEBUG (see earlier review comment).

Issue #891, opened 2026-05-08
Models: per-prompt AI model bypass inconsistent with Model Config restrictions


FExB17:
## Problem

A non-premium user who is blocked from changing models on the AI Models Config page (`/config/ai-models`, 403 on save) can still assign any model — including premium ones — via the Prompt Management page (`/config/task-prompts`) by editing or overriding a system prompt's AI Model dropdown. The two pages enforce different restrictions for the same operation.

---

## Expected

Model restrictions should be consistent across all configuration surfaces. If a user cannot select a premium model on the Model Config page, the same restriction should apply on the Prompt Management page's AI Model dropdown.

## Actual

- **Model Config page:** POST `/api/v1/config/models/defaults` returns 403 when a restricted model is selected
- **Prompt Management page:** PUT `/api/v1/prompts/{id}` (or POST for override creation) accepts any `aiModel` in metadata without checking model eligibility — the save goes through

---

## Steps to reproduce

1. Log in to web.synaplan.com as a non-premium user
2. Go to `/config/ai-models` → select a premium model → save → **403 Forbidden**
3. Go to `/config/task-prompts` → click on `(default) general-chat`
4. Change the AI Model dropdown to the same premium model
5. Save → **succeeds**, creating a custom override with the premium model

---

## Notes

- Related: #883 (the 403 error UX on Model Config)
- The prompt override path does not validate model eligibility in `PromptController::saveMetadataForPrompt`
- Fix options: either add the same eligibility check in the prompt metadata save path, or share a common `ModelEligibilityGuard` service between both controllers

##Screenshots/Logs

<img width="2547" height="1440" alt="Image" src="https://github.com/user-attachments/assets/748feb50-8bea-473b-8faf-de0611e98dff" />

<img width="2547" height="1440" alt="Image" src="https://github.com/user-attachments/assets/d7516b27-d066-414a-8574-9ce32943d527" />

orgaralf:
Architectural finding while implementing the planned shared `ModelEligibilityGuard` (see [`_devextras/planning/20250511-issues-plan.md`](https://github.com/metadist/synaplan/blob/main/_devextras/planning/20250511-issues-plan.md) and PR #908):

The current codebase **has no per-model premium attribute**. The only model-related premium gate is `App\Service\Embedding\EmbeddingModelChangeGuard` (`backend/src/Service/Embedding/EmbeddingModelChangeGuard.php`), which:

[...]

PR #908, opened 2026-05-11
fix(config): surface premium-required reason in models save toast

### LINKED ISSUE #883: AI Models Config: 403 on save shows generic error, no explanation for user
## Problem

When a user selects a PREMIUM model (e.g. Embedding/Vectorization) on the AI Models config page and clicks save, the API returns 403 Forbidden. The UI shows only a generic red toast "Failed to save model configuration" with no indication that the selected model requires a premium subscription.

---

## Expected

The UI should clearly explain WHY the save failed:
- Inline badge or tooltip on the model dropdown indicating it requires a higher plan
- Error toast should say something like "The selected Embedding model requires a Premium subscription. Please upgrade or choose a different model."
- Ideally prevent the save button from triggering at all when an ineligible model is selected

## Actual

- Red toast: "Failed to save model configuration"
- Console: `POST /api/v1/config/models/defaults` → 403 (Forbidden)
- Console error: `"Failed to save configuration: Error: integrate premium"`
- The PREMIUM badge is visible on the dropdown but there is no explanation that this blocks saving
- User has no actionable feedback — they don't know what to change

---

## Steps to reproduce

1. Log in to web.synaplan.com as a non-premium user
2. Navigate to /config/ai-models
3. Under "Embedding / Vectorization", select a model marked with PREMIUM badge
4. Click save
5. Observe: generic error toast, no explanation

---

## Notes

The backend correctly returns 403, but the frontend does not parse or display the error reason from the response body. The string "integrate premium" appears in the console error but is never surfaced to the user.

Two fixes needed:
1. **Frontend:** Parse the 403 response body and show the actual reason (e.g. "Premium required") in the toast
2. **UX:** Disable or visually gate premium models for non-premium users (grey out + "Upgrade" link) so the user never reaches a 403 in the first place

---

## Screenshots/Logs

Image

Console output:
```
POST https://web.synaplan.com/api/v1/config/models/defaults 403 (Forbidden)
Failed to save configuration: Error: integrate premium
```

orgaralf:
## Summary
Replace the generic "Failed to save model configuration" toast on `/config/ai-models` with the actual reason the backend returned (e.g. "Switching the embedding model requires an active paid subscription. Current level: NEW. Please upgrade your plan, or pick a different model.").

## Changes
- `frontend/src/services/api/httpClient.ts`:
  - New `class ApiError extends Error` exposing `status`, `code`, `details`, `debug` so consumers can branch on machine-readable codes without parsing strings.
  - Error path now throws `ApiError`. Prefers the backend's human-readable `message` over the legacy `error` code for the displayed text, so existing generic catch blocks immediately show useful text with no extra plumbing.
- `frontend/src/components/config/AIModelsConfiguration.vue`:
  - Catch block detects `ApiError` with status 403 + code `requires_premium`, wraps the backend message in the new `saveErrorPremiumRequired` i18n string with an upgrade-CTA hint.
  - Generic non-200 errors fall back to `err.message`; truly opaque errors keep the original generic toast.
- `frontend/src/i18n/{en,de}.json`: new `config.aiModels.saveErrorPremiumRequired` key in both languages.
- `frontend/tests/unit/api/httpClient.spec.ts` (new): four Vitest cases — message-preferred over code, bare-error fallback, admin debug suffix preservation, non-JSON-body synthetic-status fallback.

## Verification
- [ ] Not tested (explain)
- [x] Manual *(behavior reproduced from the issue's STR; final UX confirmation by reviewer or QA against a non-premium user account)*
- [x] Tests added/updated

Local pre-commit gate (mirrors `.github/workflows/ci.yml` frontend job):

| Step | Command | Result |
|---|---|---|
| New spec | `docker compose exec -T frontend npx vitest run tests/unit/api/httpClient.spec.ts` | 4/4 green |
| Full Vitest | `docker compose exec -T frontend npx vitest run` | 43 files / 354 tests green |

## Notes
Closes #883.

Sister issue #891 ("per-prompt AI model bypass") was originally planned to ship in the same PR, but on closer reading there is no per-model premium attribute in the current data model — the only model gate is the embedding-switch guard (`EmbeddingModelChangeGuard`), which is VECTORIZE-specific and operates on the *act of changing the active embedding model*, not on individual model identities. The prompt's `aiModel` metadata is consumed only by the chat path (`StreamController:1530`) and is not subject to that gate. A follow-up architectural discussion on whether to introduce per-model premium tagging is more appropriate than a reactive validation in `PromptController`. Linking back to #891 with that finding so the project can decide.

The new `ApiError` class is general — a follow-up could update other catch sites (memory parse 503, widget upload errors, etc.) to display the backend's chosen message instead of generic copy. Out of scope here to keep the diff focused.

## Screenshots/Logs
N/A — pure UX wiring fix; the visible output is the toast text, identical to what the backend already logs.

### DIFF
--- a/frontend/src/components/config/AIModelsConfiguration.vue
+++ b/frontend/src/components/config/AIModelsConfiguration.vue
@@ -433,6 +433,7 @@ import {
   checkModelAvailability,
 } from '@/services/api/configApi'
 import { adminEmbeddingApi, type EmbeddingGuardStatus } from '@/services/api/adminEmbeddingApi'
+import { ApiError } from '@/services/api/httpClient'
 import { useAuthStore } from '@/stores/auth'
 import type { AIModel, Capability } from '@/types/ai-models'
 import { getProviderIcon } from '@/utils/providerIcons'
@@ -923,7 +924,23 @@ const saveConfiguration = async () => {
     }
   } catch (err: unknown) {
     console.error('Failed to save configuration:', err)
-    showError(t('config.aiModels.saveError'))
+    // Issue #883: surface the actual reason the backend gave us instead of
+    // a generic "Failed to save model configuration" toast. The premium
+    // gate on `ConfigController::saveDefaultModels` returns a structured
+    // 403 `{ error: 'requires_premium', message: 'Switching the embedding
+    // model requires an active paid subscription. Current level: NEW.', ... }`
+    // and `httpClient.ApiError` now exposes both the message and the code.
+    if (err instanceof ApiError && 403 === err.status) {
+      const reason =
+        'requires_premium' === err.code
+          ? t('config.aiModels.saveErrorPremiumRequired', { reason: err.message })
+          : err.message
+      showError(reason)
+    } else if (err instanceof Error && err.message) {
+      showError(err.message)
+    } else {
+      showError(t('config.aiModels.saveError'))
+    }
   } finally {
     saving.value = false
   }
--- a/frontend/src/i18n/de.json
+++ b/frontend/src/i18n/de.json
@@ -891,6 +891,7 @@
       "noModelsAvailable": "Keine Modelle für diesen Verwendungszweck verfügbar",
       "saveSuccess": "Modellkonfiguration gespeichert",
       "saveError": "Modellkonfiguration konnte nicht gespeichert werden",
+      "saveErrorPremiumRequired": "{reason} Bitte aktualisiere deinen Tarif oder wähle ein anderes Modell.",
       "tableName": "Name",
       "tableService": "Anbieter",
       "tableRating": "Bewertung",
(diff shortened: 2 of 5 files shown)


FExB17 (approved):
LGTM. Clean introduction of `ApiError`, backward-compatible, good test coverage.

orgaralf:
Copilot's point about mixed-language toasts is **valid** — German users currently see "Switching the embedding model requires an active paid subscription. Current level: NEW. Bitte aktualisiere deinen Tarif oder wähle ein anderes Modell." (English reason + German CTA). Properly fixing it requires:

1. Backend: stop sending a hardcoded English `message`, send only the structured `{ error: 'requires_premium', capability, currentLevel }` payload.
2. Frontend: build the toast text purely from i18n + those structured fields (`config.aiModels.errors.requires_premium.{free,new,…}`).

That's a coordinated backend + frontend change which doesn't fit the focused scope of this PR (frontend-only, surface what the backend already returns). Tracking as a follow-up — happy to pick it up after merge, or it can become a small scoped backend PR + matching frontend update.

Until then the wins from this PR still hold:
- English-locale users get a useful actionable toast instead of "Failed to save model configuration".
- The new `ApiError` class is general and lets the next caller localize without guessing — when the backend stops sending hardcoded text, only the consumer's branch needs updating.

PR #909, opened 2026-05-11
fix(error): downgrade transient chunk-load / network errors to a toast

### LINKED ISSUE #897: Mobile(Browser): Full-screen error on file upload send and after background resume
## Problem

On mobile (Android), a full-screen "Etwas ist schiefgelaufen" error replaces the entire UI in at least two scenarios: (1) sending a message with an uploaded file attachment, and (2) returning to the app tab after it has been in the background for a few minutes.

---

## Expected

File upload + send should work normally and display the AI response. Returning to the app from background should restore the previous view or silently reconnect.

## Actual

A full-screen `ErrorView` is shown with "Etwas ist schiefgelaufen" ("Something Went Wrong"), requiring a manual page reload to continue using the app.

---

## Steps to reproduce

**Scenario A — File upload + send:**
1. Open the app on a mobile device (Android, tested in Chrome, not PWA).
2. Open any chat.
3. Upload a file via the attachment button.
4. Tap send.
5. Observe: full-screen error instead of the AI response.

**Scenario B — Background resume:**
1. Open the app on a mobile device.
2. Use the app normally (e.g. open a chat).
3. Switch to the home screen or another app.
4. Wait a few minutes or longer.
5. Return to the app tab.
6. Observe: full-screen error instead of the previous view.

---

## Notes

Both scenarios trigger the same full-screen `ErrorView` via `installGlobalErrorHandlers.ts`, which catches **all** unhandled promise rejections and uncaught errors and pipes them into `useGlobalErrorStore().setError()`.

**Scenario A likely path:** `handleSendMessage` in `ChatView.vue` (~line 1196) is async and uses a dynamic import (`import('@/services/filesService')`) plus `filesService.getFileContent(fileId)` for each attached file. If any async step in this chain or in the subsequent `streamAIResponse()` throws on mobile (e.g. network hiccup, chunk load failure, SSE connection error), the unhandled rejection hits the global handler and triggers the fatal error screen.

**Scenario B likely path:** After background suspension, the browser may evict resources. On resume, stale JS chunks (`import()` failures), expired SSE tokens, or other async operations may throw without a local catch. The SW (`public/sw.js`) aggressively deletes all caches on activate and has no fetch handler, which could compound chunk-load failures after an update.

**Root cause pattern:** The global error handler (`installGlobalErrorHandlers.ts`) is too aggressive — it treats every unhandled rejection as fatal and shows the full-screen error overlay. Many of these errors (network timeouts, chunk loads, SSE reconnection failures) should be handled gracefully with a retry or a non-fatal toast notification instead.

**Possible fixes:**
- Add more patterns to `IGNORED_PATTERNS` for transient network/chunk errors
- Wrap `handleSendMessage` and `streamAIResponse` in a top-level try/catch with user-friendly error toast instead of relying on the global handler
- Add chunk-load retry logic (common Vite pattern for `import()` failures)
- Downgrade certain rejection classes to non-fatal toasts rather than full-screen error

related: `frontend/src/utils/installGlobalErrorHandlers.ts`, `frontend/src/views/ChatView.vue` (~line 1196), `frontend/src/views/ErrorView.vue`

---

## Screenshots/Logs

Image

orgaralf:
## Summary
Stop the global error handler from showing the full-screen "Etwas ist schiefgelaufen" view for transient browser failures (chunk-load on mobile after the tab was backgrounded, dynamic-import failures on flaky cellular, generic `Failed to fetch`, SSE connection blips). Surface a non-fatal warning toast instead so the chat / view stays usable and the user can simply retry.

## Changes
- `frontend/src/utils/installGlobalErrorHandlers.ts`:
  - Add `TRANSIENT_PATTERNS` covering Vite chunk-load (`Loading chunk \d+ failed`, `ChunkLoadError`, `Failed to fetch dynamically imported module`, `Importing a module script failed`), generic network errors (`Failed to fetch`, Safari's `Load failed`, `NetworkError when attempting to fetch resource`, `TypeError: NetworkError`) and SSE (`EventSource ... failed`).
  - New `isTransient()` predicate + `reportTransient()` helper that pushes a `warning` toast through `useNotification()` (module-scoped, no Vue context required) and console.warns the source.
  - Wire the predicate into all three handler channels (Vue `errorHandler`, window `error`, window `unhandledrejection`) BEFORE the `store.setError` fall-through, so transient matches never reach the full-screen view.
- `frontend/tests/unit/utils/installGlobalErrorHandlers.spec.ts`:
  - 5 transient cases covered via `it.each` (chunk load, dynamic import, ChunkLoadError, Failed to fetch, Safari Load failed, EventSource failure).
  - 1 explicit Scenario-A unhandled-rejection repro mirroring `handleSendMessage`'s dynamic `import('@/services/filesService')` failing.
  - 1 negative control asserting that a real `TypeError` is still raised to the full-screen view (so downgrading transient errors doesn't silently mask real bugs).
  - Pre-existing 7 cases unchanged.
  - Per-test toast queue reset added in `beforeEach` so transient assertions don't bleed across cases.

## Verification
- [ ] Not tested (explain)
- [x] Manual *(real-device sweep on Android delegated to reviewer / QA — environment outside CI; the unit-test cases reproduce the exact production messages from the issue's "likely path" notes)*
- [x] Tests added/updated

Local pre-commit gate (mirrors `.github/workflows/ci.yml` frontend job, including the build step):

| Step | Command | Result |
|---|---|---|
| Spec | `docker compose exec -T frontend npx vitest run tests/unit/utils/installGlobalErrorHandlers.spec.ts` | 15/15 green |
| Full Vitest | `docker compose exec -T frontend npx vitest run` | 42 files / 358 tests green |
| Build | `docker compose exec -T frontend npm run build` | green (1.64s, only pre-existing chunk-size warnings) |

## Notes
Closes #897.

The `sw.js` cache nuke + missing `fetch` handler from the issue's "likely path" notes is intentionally **out of scope** for this PR — it's a much bigger production change and the toast-downgrade alone covers both Scenario A (file upload + send) and Scenario B (background resume) on the data we have. Can be revisited as a follow-up if real-device QA still hits a fatal screen.

The chunk-load retry-once-then-reload helper around the dynamic `import('@/services/filesService')` in `ChatView.vue::handleSendMessage` was also considered. With the global handler now downgrading the same error class, the user sees a warning toast instead of a fatal screen and can simply re-send — the retry-once helper became redundant. Skipped to keep the diff small and the surface tight.

## Screenshots/Logs
N/A — visual change is "the full-screen ErrorView no longer appears for these error classes". Reproducing in CI requires real mobile network conditions; the spec captures the symptomatic messages directly.

### DIFF
--- a/frontend/src/utils/installGlobalErrorHandlers.ts
+++ b/frontend/src/utils/installGlobalErrorHandlers.ts
@@ -1,6 +1,7 @@
 import type { App } from 'vue'
 import { useGlobalErrorStore, type GlobalErrorPayload } from '@/stores/globalError'
 import { getErrorMessage } from '@/utils/errorMessage'
+import { useNotification } from '@/composables/useNotification'
 
 /**
  * Patterns we deliberately ignore. These are typically benign noise produced by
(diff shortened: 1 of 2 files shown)


FExB17 (approved):
LGTM. Patterns are conservative, negative control test ensures real errors aren't swallowed. One minor thought: the hardcoded English toast `"Connection hiccup. Please try again."` should probably go through i18n — but that's a follow-up nit, not a blocker.

PR #926, opened 2026-05-12
fix(mobile): scroll-aware header with floating FAB, fix keyboard viewport

### LINKED ISSUE #896: Mobile: Header with menu button scrolls out of view in chat, large bottom gap
## Problem

On mobile, the app header (burger menu button) disappears when the user scrolls down in the chat view and only reappears when scrolling all the way back to the top, making the sidebar menu effectively unreachable during a conversation. Additionally, a noticeable gap appears between the chat input and the bottom edge of the display.

---

## Expected

The header with the burger menu button should remain accessible at all times during chat, regardless of scroll position. The chat input should sit flush against the bottom of the viewport (respecting only the safe-area inset for the home indicator).

## Actual

1. The header (containing the burger menu button) vanishes when the user scrolls to the bottom of a chat conversation. To access the menu, the user must scroll all the way back to the top of the chat history.
2. A large gap appears between the chat input area and the bottom edge of the display.

---

## Steps to reproduce

1. Open the app on a mobile device  in a  browser.
2. Navigate to a chat with enough messages to require scrolling.
3. Scroll down to the bottom of the chat.
4. Observe: the header with the burger menu button is no longer visible.
5. Try to open the sidebar menu — it is not possible without scrolling back to the top.
6. Observe the gap between the chat input and the bottom display edge.

---

## Notes

- `MainLayout.vue` uses `h-dvh` on the root container and places `<Header />` in normal flow (not `sticky` or `fixed`) above the scrollable `<main>` region.
- The `<main>` element uses `flex-1 min-h-0 overflow-y-auto`, so in theory the header should not scroll away — the content scrolls inside `<main>`.
- Likely cause: mobile browser chrome (address bar hide/show) interacts with `dvh` units, potentially causing the outer flex container to exceed the visual viewport and introducing body-level scroll that carries the header offscreen.
- The bottom gap is likely related to `env(safe-area-inset-bottom)` padding on `ChatInput.vue` combined with the viewport resize from browser chrome changes.
- Consider making the `<Header>` `sticky top-0` or `fixed` on mobile to keep the burger menu always reachable.

related: PR #832 (dvh fix for mobile Safari)

---

## Screenshots/Logs

Image

FExB17:
## Summary

Fixes mobile UX: the header (burger menu) scrolling out of view, a ~1cm gap below the chat input, and keyboard covering chat content.

## Changes

- **Body scroll lock:** `html, body { overflow: hidden; overscroll-behavior: none }` prevents document-level scroll regardless of dvh fluctuations from browser chrome changes.
- **MainLayout:** `overflow-hidden` on outer container, `overscroll-contain` on `<main>`.
- **In-flow header bar:** Header sits in normal document flow (above `<main>`), no overlap with content. Collapses via `max-height` + `overflow: hidden` as user scrolls past the first 36px.
- **Floating FAB:** A round 42px button fades in proportionally when the header bar collapses, providing menu access at all scroll positions. Hidden when sidebar is open.
- **`useHeaderVisibility` composable:** Scroll-position-based progress (0–1) shared across all views. Virtual keyboard guard prevents false triggers.
- **ToolsView:** Scroll tracking added so header visibility works in Tools/Mail-Handler/Doc-Summary views.
- **ChatView:** `overscroll-contain` on messages container. `sync()` after auto-scroll-to-bottom keeps header hidden until user scrolls to top. `visualViewport` resize handler keeps chat scrollable when keyboard opens.
- **Viewport meta:** Added `interactive-widget=resizes-content` for proper keyboard-aware layout.

## What's NOT affected

- **Desktop:** Header wrapper has `md:hidden`, FAB has `md:hidden`. No changes visible.
- **iOS:** `env(safe-area-inset-top)` still respected via existing inline style.
- **All modals/dropdowns:** Use `<Teleport>` or `position: fixed`, never clipped.

## Compatibility

All CSS/JS features used are universally supported (no CSS grid `0fr` trick — uses `max-height` instead). `interactive-widget` and `visualViewport` are progressive enhancements with guards.

## Verification

- [x] vue-tsc clean
- [x] ESLint + Prettier clean
- [x] 49 files / 403 tests green
- [x] Manual: tested on mobile — header collapses on scroll, FAB appears, no overlap with titles, keyboard doesn't cover chat

Closes #896

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -2,7 +2,7 @@
 <html lang="en" class="design-v2">
   <head>
     <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover, interactive-widget=resizes-content" />
 
     <!-- Favicon for modern browsers (SVG) -->
     <link rel="icon" type="image/svg+xml" href="/single_bird.svg" />
--- a/frontend/src/components/Header.vue
+++ b/frontend/src/components/Header.vue
@@ -1,25 +1,42 @@
 <template>
-  <header
-    :class="[
-      'fixed top-0 inset-x-0 z-50 bg-header border-b border-black/[0.04] dark:border-white/[0.04] md:hidden',
-      'transition-transform duration-200 ease-out',
-      hidden ? '-translate-y-full' : 'translate-y-0',
-    ]"
+  <div
+    class="md:hidden overflow-hidden"
+    :style="{ maxHeight: `${(1 - progress) * 100}px` }"
     data-testid="comp-app-header"
-    :style="{ paddingTop: 'env(safe-area-inset-top)' }"
   >
-    <div class="flex items-center px-3 py-1.5" data-testid="section-header-bar">
-      <button
-        type="button"
-        class="btn-sidebar-burger"
-        aria-label="Toggle sidebar"
-        data-testid="btn-sidebar-toggle"
-        @click="sidebarStore.toggleMobile()"
-      >
-        <Bars3Icon class="w-6 h-6" aria-hidden="true" />
-      </button>
-    </div>
-  </header>
+    <header
+      class="bg-header border-b border-black/[0.04] dark:border-white/[0.04]"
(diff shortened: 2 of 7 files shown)


orgaralf (approved):
question of taste, but I also like my menu fixed!

PR #930, opened 2026-05-13
fix(billing): record FILE_ANALYSIS once per (user, file) and only on actual analysis

### LINKED ISSUE #887: Billing: FILE_ANALYSIS counted on upload (not analysis) and double-counted on RAG retry
## Problem

`FILE_ANALYSIS` usage is recorded at the wrong point in the lifecycle and can be double-counted in certain RAG flows.

---

## Expected

1. `FILE_ANALYSIS` should be counted when a file is actually **analysed** (text extracted / vectorized), not merely uploaded.
2. A single file upload should produce exactly **one** `FILE_ANALYSIS` record, regardless of retry or async processing.
3. Uploading a file in chat but never sending the message should **not** count as a file analysis.

### 1. Chat upload counts on upload, not on analysis

`MessageController::uploadFileForChat()` calls `recordUsage($user, 'FILE_ANALYSIS')` immediately after storing the file (line 566), before the file is ever used in a message. If the user uploads a file but never sends the message, `FILE_ANALYSIS` is still consumed.

Non-audio files are not even extracted at upload time — extraction is deferred to `MessagePreProcessor` during the stream. Yet the usage is already billed.

### 2. RAG document upload can double-count

`FileUploadService` has two methods that both call `recordUsage(…, 'FILE_ANALYSIS')`:

- **`processSingleUpload()`** — records after successful vectorization (process_level `vectorize` or `full`), even if vectorization failed or extracted text was empty.
- **`processFile()`** — records on two branches (empty text → mark vectorized, or successful vectorization).

When `processSingleUpload()` runs with `vectorize`/`full` but vectorization **fails** (file stays at status `extracted`), the user can retry via `POST /api/v1/files/{id}/process`. If `processFile()` then succeeds, **both methods have called `recordUsage`** → **2 increments** for one file.

Similarly, if `processSingleUpload()` records on the empty-text path and the client later calls `/process`, a second record is created.

### 3. Widget upload is correct

`WidgetPublicController::upload()` records exactly once per successful upload (charged to widget owner). No double-count path.

---

### Double-count (RAG):
1. Upload a document via `POST /api/v1/files/upload` with `process_level=vectorize` while Qdrant is down or vectorization fails
2. Check `BUSELOG`: one `FILE_ANALYSIS` entry exists
3. Call `POST /api/v1/files/{id}/process` — vectorization succeeds
4. Check `BUSELOG`: **two** `FILE_ANALYSIS` entries for the same file

### Upload-without-send (Chat):
1. Upload a file via `POST /api/v1/messages/upload-file`
2. Do not send any message
3. Check `BUSELOG`: one `FILE_ANALYSIS` entry exists despite no analysis happening

---

## Notes

- related: Issue #886 (cost calculation gaps), PR #637 (token/cost tracking)
- All `FILE_ANALYSIS` recording locations:
  - `MessageController::uploadFileForChat()` — 1x per chat upload (source: `WEB`)
  - `FileUploadService::processSingleUpload()` — 1x per RAG upload with vectorize/full (source: `WEB`)
  - `FileUploadService::processFile()` — 1x per async processing (source: `WEB_ASYNC`), two branches
  - `WidgetPublicController::upload()` — 1x per widget upload (source: `WIDGET`)
- `StreamController`, `MessagePreProcessor`, `ChatHandler` do **not** record `FILE_ANALYSIS`
- Suggested fix: move chat `FILE_ANALYSIS` recording into the stream processing (when file content is actually used), and add a guard in `processFile()` to skip if `processSingleUpload()` already recorded for the same file

---

## Screenshots/Logs

| Flow | User action | FILE_ANALYSIS records | Correct? |
|------|------------|----------------------|----------|
| Chat upload + send | Upload file, send message | 1 (at upload) | ⚠️ Counted too early |
| Chat upload, no send | Upload file, abandon | 1 (at upload) | ❌ Should be 0 |
| RAG upload (success) | Upload + vectorize | 1 | ✅ |
| RAG upload (fail) + retry | Upload fails vectorize, then /process | 2 | ❌ Should be 1 |
| Widget upload | Visitor uploads file | 1 (owner) | ✅ |

orgaralf:
## Summary
Fix two `FILE_ANALYSIS` lifecycle bugs from #887: the chat upload billed users for an analysis that never happened (when they uploaded a file but didn't send the message), and the RAG retry path double-counted when `processSingleUpload()` and `processFile()` both fired against the same File entity.

## Changes
- `backend/src/Service/RateLimitService.php`:
  - New `recordFileAnalysisOnce(User $user, int $fileId, array $metadata): bool` helper. Idempotent on `(BUSERID, BACTION='FILE_ANALYSIS', BMETADATA->file_id)` via a `JSON_EXTRACT` lookup on the existing JSON-typed `BMETADATA` column. Returns `true` if a row was written, `false` if the (user, file) pair already had one. `file_id <= 0` falls through to legacy `recordUsage()` for callers without a file id.
- `backend/src/Controller/MessageController.php` (`uploadFileForChat`):
  - Remove the unconditional `recordUsage('FILE_ANALYSIS')` after upload. The pre-upload `checkLimit('FILE_ANALYSIS')` gate stays in place so users at quota can't keep uploading; only the BUSELOG accounting moves.
- `backend/src/Service/Message/MessagePreProcessor.php`:
  - Inject `RateLimitService` + `UserRepository`.
  - New private helper `billFileAnalysis()` that calls `recordFileAnalysisOnce` and swallows failures (best-effort accounting must never block a streamed message).
  - Wire it into the three success branches of `processMessageFile()` (document via Tika, audio via Whisper/STT, image via Vision AI) **and** the early-return path where text was already extracted at upload time. Re-runs of the preprocessor (retry, HMR replay) are no-ops thanks to the dedup.
- `backend/src/Service/File/FileUploadService.php`:
  - `processSingleUpload()` and both branches of `processFile()` switch from `recordUsage` to `recordFileAnalysisOnce`. A vectorize retry after Qdrant comes back now collapses to a single BUSELOG row.
- `backend/tests/Service/RateLimitServiceFileAnalysisOnceTest.php` **(new)** — 6 cases covering: first-call writes a row, second-call no-op, distinct file ids independent, `file_id=0` fallback, negative-id fallback, file_id forced into metadata even when caller omits it.
- `backend/tests/Unit/MessagePreProcessorTest.php` — constructor sites updated for the two new dependencies (3 sites). Existing assertions unchanged.
- `_devextras/planning/20250513-prio1-round2.md` **(new)** — round-2 plan doc (next 5 `prio:1`) shipped in this PR for traceability.

## Verification
- [ ] Not tested (explain)
- [x] Manual *(behaviour reproduced from the issue's STR; the four-row table in the issue's "Screenshots/Logs" is now satisfied — see test plan)*
- [x] Tests added/updated

Local pre-commit gate (mirrors `.github/workflows/ci.yml` backend job):

| Step | Command | Result |
|---|---|---|
| New + adjacent tests | `php bin/phpunit tests/Service/RateLimitServiceFileAnalysisOnceTest.php tests/Unit/MessagePreProcessorTest.php tests/Service/RateLimitServiceUnifiedTest.php tests/Service/File/FileUploadServiceCheckUploadTest.php` | 28/28 green, 88 assertions |
| Full PHPUnit | `php bin/phpunit` | **1492 / 1492 green**, 5119 assertions |
| PHPStan | `make -C backend phpstan` | clean (624 files) |

The four-row table in the issue (`Chat upload + send` / `Chat upload, no send` / `RAG upload (success)` / `RAG upload (fail) + retry` / `Widget upload`) now produces:

| Flow | Before | After |
|---|---|---|
| Chat upload + send | 1 (at upload, wrong moment) | **1** (at analysis, correct moment) |
| Chat upload, no send | 1 ❌ | **0** ✅ |
| RAG upload (success) | 1 | 1 |
| RAG upload (fail) + retry | 2 ❌ | **1** ✅ |
| Widget upload | 1 (owner) | 1 (owner) — unchanged |

## Notes
Closes #887. Companion to the round-2 hunt described in `_devextras/planning/20250513-prio1-round2.md` — next four PRs in this series will land separately.

`recordFileAnalysisOnce` is intentionally scoped to FILE_ANALYSIS even though the dedup pattern would generalise. Other actions (MESSAGES, EMBEDDINGS, IMAGES) genuinely DO want one BUSELOG row per call site since the same chat turn can legitimately produce multiple of them. Generalising into `recordUsageOnce` would be premature.

## Screenshots/Logs
N/A — pure backend accounting fix; the visible difference is row counts in `BUSELOG`, captured exactly by the new PHPUnit cases.

### DIFF
--- a/backend/src/Controller/MessageController.php
+++ b/backend/src/Controller/MessageController.php
@@ -575,12 +575,21 @@ class MessageController extends AbstractController
                 'status' => $messageFile->getStatus(),
             ]);
 
-            // Record FILE_ANALYSIS usage for statistics
-            $this->rateLimitService->recordUsage($user, 'FILE_ANALYSIS', [
-                'file_id' => $messageFile->getId(),
-                'filename' => $uploadedFile->getClientOriginalName(),
-                'source' => 'WEB',
-            ]);
+            // FILE_ANALYSIS recording is INTENTIONALLY deferred to
+            // MessagePreProcessor::processMessageFile() — a chat upload
+            // that the user never sends should not be billed (issue #887).
+            // The pre-upload `checkLimit('FILE_ANALYSIS')` above still
+            // rejects users who have already exhausted their quota, so the
+            // limit gate stays in place; we just no longer write a
+            // BUSELOG row at the point a file is staged.
+            //
+            // For audio files we extracted synchronously above, recording
+            // also happens via MessagePreProcessor: when the message is
+            // streamed (or when an audio-only upload triggers an immediate
+            // analysis through the standard pipeline), the file id is
+            // looked up by the preprocessor and routed through
+            // RateLimitService::recordFileAnalysisOnce(), which dedups so
+            // the same file never bills twice.
 
             $response = [
                 'success' => true,
--- a/_devextras/planning/20250513-prio1-round2.md
+++ b/_devextras/planning/20250513-prio1-round2.md
@@ -1,0 +1,70 @@
+# `prio:1` Hunt — Round 2 (2026-05-13)
+
+> Source: <https://github.com/metadist/synaplan/issues> filtered by `is:issue state:open label:"prio:1"`
+> Snapshot taken **2026-05-13 09:40 UTC+2**, 7 open `prio:1` issues remain after Round 1.
+> Companion to [`20250511-issues-plan.md`](./20250511-issues-plan.md).
+
+## Inventory — 7 open `prio:1` issues
+
+| # | Title | Round-2 status |
+|---|---|---|
+| #439 | Memories: "Create with AI" fails with 503 | **deferred** — environment-dependent (Qdrant / AI provider availability), not safely closeable from static analysis |
+| #626 | Mail: Generated video not displayed on platform | **deferred** — needs cross-channel runtime verification |
+| #839 | Statistics: Inconsistent request counts and misleading subscription status | **target #3** |
+| #856 | Subscription: payment_failed has no UI surface and the past_due status badge is broken | **target #2** |
+| #886 | Billing: cost calculation gaps + rate-limiting still uses request count | **target #4 / #5** (split into sub-PRs) |
+| #887 | Billing: FILE_ANALYSIS counted on upload (not analysis) and double-counted on RAG retry | **target #1** |
+| #891 | Models: per-prompt AI model bypass inconsistent with Model Config restrictions | **blocked** — awaits product decision (architectural comment posted on GH) |
+
+## The next 5 to attack
+
+Order chosen for: smallest blast radius first → testable in PHPUnit alone → progressively wider reach.
+
+| Order | Issue | Branch | Scope |
+|---|---|---|---|
+| **1** | **#887** | `fix/887-file-analysis-lifecycle` | Move chat `FILE_ANALYSIS` recording out of `MessageController::uploadFileForChat` into the actual analysis point in `MessagePreProcessor`. Add idempotency guard to `FileUploadService::processFile`. Backend-only, fully PHPUnit-testable. |
+| **2** | **#856** | `fix/856-subscription-past-due-ui` | Backend: clear `payment_failed` flag in `handlePaymentSucceeded` and `handleSubscriptionUpdated` when status returns to `active/trialing`. Frontend: dedicated warning section above current-plan; status-text resolver maps snake_case → camelCase i18n keys. E2E spec asserts the new behaviour. |
+| **3** | **#839** | `fix/839-statistics-counters-and-status` | Sub-fixes only (the broader cost-budget gate is in #886): label the three counters distinctly in i18n, fix "Inactive" → "Free" for active free-plan users, fix `phone_verified` arg in `StreamController`. NO request-gate wiring here. |
+| **4** | **#886** sub-PR (a) | `fix/886a-image-pricing-mode` | Set `pricing_mode: per_image` on every image model in `ModelCatalog.php`; unit-test cost calculation flow in `RateLimitService::recordUsage`. |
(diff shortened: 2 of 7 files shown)


backend/src/Controller/MessageController.php
```diff
@@ -575,12 +575,21 @@ public function uploadFileForChat(
...
+            // limit gate stays in place; we just no longer write a
+            // BUSELOG row at the point a file is staged.
+            //
+            // For audio files we extracted synchronously above, recording
+            // also happens via MessagePreProcessor: when the message is
+            // streamed (or when an audio-only upload triggers an immediate
+            // analysis through the standard pipeline), the file id is
+            // looked up by the preprocessor and routed through
+            // RateLimitService::recordFileAnalysisOnce(), which dedups so
+            // the same file never bills twice.
```
FExB17, line 592: This 14-line comment block replaces 5 lines of removed code. Most of the explanation is already covered by the docblock on `recordFileAnalysisOnce()` and the PR description — repeating it inline will rot when the design evolves.

Suggestion: trim to 2–3 lines, e.g.:
```php
// FILE_ANALYSIS billing deferred to MessagePreProcessor::processMessageFile()
// so uploads that are never sent don't get billed (issue #887).
// The checkLimit() gate above still enforces the quota.
```

FExB17 (commented):
Verified: both bugs from #887 are correctly fixed.

1. **Chat upload-without-send** no longer bills — billing is deferred to `MessagePreProcessor::processMessageFile()` where the actual analysis happens.
2. **RAG double-count on retry** is resolved via `recordFileAnalysisOnce()` dedup on `(user_id, file_id)`.
3. Widget path is correctly left unchanged.
4. All four branches in `processMessageFile()` (reuse_extracted, document, audio, image) are wired to `billFileAnalysis()`.
5. Error/skip paths (Whisper unavailable, Vision AI failure) correctly do NOT bill — consistent with "only bill when analysis actually ran."

One comment below.

FExB17 (approved):

PR #925, opened 2026-05-12
Fix batch 2: chat memory leaks, MEDIAMAKER live render & WhatsApp poll cascade (#583, #615, #624, #625, #633)

cristiangrxs:
## Summary
Second batch of GitHub issue fixes building on `github-issues-fixing` (merged via #910). Closes five user-reported issues across chat streaming, memory extraction, and the WhatsApp webhook pipeline.

Closes #583
Closes #615
Closes #624
Closes #625
Closes #633

### #583 — Chat: surface TTS model separately from chat model in voice replies
- `applyAssistantChatModelFooter()` (frontend `chatModelBadge.ts`) now resolves the chat model and the audio/TTS model as independent badges instead of overwriting one with the other.
- New `chatBadgeLabel()` helper localizes the badge ("Chat Model" / "Audio Model" / "Sorting Model") from the resolved capability.
- Backend `StreamController` persists `ai_audio_provider` / `ai_audio_model` / `ai_audio_model_id` meta on the outgoing message when TTS runs, so a page reload sees the same labels as the live view.

### #615 — Chat: load memories on email + generic webhook channels
- `ChatHandler::handle()` (non-streaming path used by email + generic API webhook) was missing memory and feedback context entirely. Refactored both `handle()` and `handleStream()` to share four new helpers:
  - `createSharedVectorResolver()` — lazy, single-call user-query embedding cache.
  - `loadMemoriesContext()` / `loadFeedbackContext()` — produce prompt fragments + emit `memories_loaded` / `feedback_loaded` SSE events.
  - `dispatchMemoryExtractionAsync()` — fire-and-forget memory extraction on the messenger bus, respecting the `PERF.V2_PIPELINE_ENABLED` kill-switch.
- All channels (Web UI / email / generic webhook) now use and contribute to long-term memory consistently.

### #624 — Chat: model badge label inconsistent between live view and page reload
- Live MEDIAMAKER audio used to show "Chat Model" → flip to "Audio Model" after refresh, and the "Again" dropdown jumped from CHAT to TEXT2SOUND capability.
- Backend extracts `persistOriginalMediaMeta()` and includes `originalTopic` + `originalMediaType` in the `complete` SSE payload for both streaming and non-streaming paths, so the frontend can resolve `mediaHint` correctly before the audio file even lands.

### #625 — Chat: MEDIAMAKER audio player not displayed in live SSE view, only after refresh
[...]

PR #934, opened 2026-05-13
refactor(chat-handler): apply PR #925 Copilot review follow-ups

cristiangrxs:
## Summary
Cleanup follow-up to the already-merged PR #925: address three Copilot review findings on \`ChatHandler\` without changing user-visible behaviour, only making the contract and the log output match what the implementation actually does.

## Changes
- \`MessageHandlerInterface::handle()\` now declares \`array $options = []\`, mirroring the existing \`ChatHandler::handle()\` implementation and the existing \`handleStream()\` symmetry. \`MediaGenerationHandler\` and \`FileAnalysisHandler\` get the same trailing parameter so PHP's LSP enforcement can't bite in production. \`InferenceRouter::route()\` gains the same optional pass-through so the email/webhook path can carry channel/\`disable_memories\` flags down to handlers — same shape as \`routeStream()\`.
- \`loadMemoriesContext()\` previously returned a single combined \`disabled\` flag that conflated "caller opted out" with "user disabled memories in profile settings". It now returns two precise booleans (\`disabledByRequest\`, \`disabledByUser\`).
- \`dispatchMemoryExtractionAsync()\` used to log _"Skipping memory extraction for widget request"_ even when the actual reason was the user's profile setting. It now takes both flags and emits a distinct log line per reason so operators don't have to spelunk through user records to tell the cases apart.
- \`loadFeedbackContext()\` renamed its locals to match the new vocabulary (\`disabledByRequest\` / \`disabledByUser\`) — same behaviour, clearer intent.
- New focused unit test \`testHandleLogsUserSettingReasonWhenMemoriesDisabledByUser\` pins the log differentiation: a non-widget caller (e.g. email channel) with a user whose memories are disabled must emit the user-setting log line and must NOT emit the widget log line.
- \`backend/phpstan-baseline.neon\` regenerated to absorb the mock-method-not-found counts introduced by the new test (purely cosmetic — same kinds of pre-existing ignore entries the rest of the test suite already has).

## Verification
- [x] Tests added/updated
- [x] Manual
  - \`make -C backend lint\` — clean
  - \`make -C backend phpstan\` — no errors
  - \`make -C backend test\` — 1487 tests / 5320 assertions / 0 failures
  - \`make -C frontend lint\` — clean
  - \`docker compose exec frontend npm run check:types\` — clean
  - \`make -C frontend test\` — 426 specs / 0 failures

## Notes
- Follow-up to merged PR #925.
- Addresses the three Copilot suggestions left on that PR:
  - PR #925 \`ChatHandler.php:79\` — interface contract / \`$options\` signature drift.
  - PR #925 \`ChatHandler.php:123\` — misleading combined \`disabled\` flag.
  - PR #925 \`ChatHandler.php:2142\` — misleading "widget request" log when the reason is the user setting.
- No user-visible behaviour change — same code paths, same routing, just precise logs and a cleaner interface.

## Screenshots/Logs
n/a (refactor + logging precision; no UI surface)


yusuf57rl (approved):
Tremendous

PR #947, opened 2026-05-18
fix: batch five prio:1 issues — memories AI fallback, mail-channel media, prompt model gate, MessageClassifier routing, SafeModelChange safety

### LINKED ISSUE #439: Memories: "Create with AI" fails with 503; AI structuring unavailable, saved as note

### LINKED ISSUE #626: Mail: Generated video not displayed on platform – only description shown
## Problem

When a video generation request is sent via email (smart+fexb17@synaplan.net), the platform receives and processes the request but does not display the generated video in the chat – only the text description/prompt is shown.

---

## Expected

The generated video (or a playable video element/link) should be visible in the chat message on the platform after the video generation completes.

## Actual

Only the text description of the generated video is displayed in the chat. No video player, thumbnail, or download link is shown.

---

## Steps to reproduce

1. Send an email to smart+fexb17@synaplan.net with subject "video test" and body requesting a video generation (e.g. "Generiere ein Video eines Jaguar in der Luft mit Auspuff und vier Beinen").
2. Wait for the AI to process and respond (Chat Model: veo-3.1-generate-preview).
3. Open the chat on web.synaplan.com.
4. Observe that the response only contains the video description text, not the actual video.

---

## Notes

- Chat Model used: `veo-3.1-generate-preview`
- The AI response contains a detailed German-language video description but no embedded video or media element.
- Must be verified for all media types (video, image, audio) – check whether generated media is correctly displayed/embedded across all generation models and channels (email, web, widget).

---

## Screenshots/Logs

Image

### LINKED ISSUE #891: Models: per-prompt AI model bypass inconsistent with Model Config restrictions

### LINKED ISSUE #948: Embedding: model switch is fundamentally broken (dimension mismatch, no reindex, no rollback)
## Problem

Switching the VECTORIZE embedding model via the admin UI leaves the system in a broken state where live memory storage silently fails (Qdrant 400), the re-vectorize job never actually re-embeds user memories, and there is no rollback if the job fails.

---

## Expected

After switching the embedding model:
1. Existing memories/documents are correctly re-embedded with the new model
2. New memories stored during live chat use compatible vector dimensions
3. If the switch fails (no API key, provider down), the system rolls back to the previous working state
4. The user/admin sees clear feedback about success or failure

## Actual

Multiple compounding failures:

1. **Live path has no dimension normalization** — `UserMemoryService::storeInQdrant()` sends raw embeddings to Qdrant without checking/coercing dimensions. If the model produces 1536-dim vectors but the collection expects 1024, Qdrant returns HTTP 400 and the memory is silently lost (logged as warning in worker, no user feedback).

2. **Memories reindex is a no-op** — `EmbeddingReindexService::reindexMemories()` calls `scrollMemories(0, null, 50000)` which filters on `user_id === 0`. Real user memories (user_id > 0) are never scrolled, never re-embedded. The "22 chunks processed" shown in the UI are only documents + synapse topics.

3. **No collection recreation for memories/documents** — Only `recreateSynapseCollection(dim)` exists. There is no equivalent for `user_memories` or `user_documents`. The collection dimension is fixed at creation time (1024 via `QdrantClientDirect::DEFAULT_VECTOR_DIM`) and never changes, regardless of which model is configured.

4. **BCONFIG updated before reindex, no rollback** — `AdminEmbeddingController::switch()` flushes the new model to BCONFIG before dispatching the reindex job. If the job fails (100% chunk failures, provider unavailable), the config stays on the broken model. All subsequent live writes fail.

5. **No API key validation before switch** — Neither the switch endpoint nor the cost-estimate endpoint checks if the target model's provider has a configured API key. User can switch to Cloudflare without a key → reindex queued → all embeddings fail → config stuck on broken model.

6. **Run with 100% failures reports "completed"** — `ReVectorizeMessageHandler` sets status `completed` whenever `execute()` returns without throwing. Per-chunk failures (including 0 processed / 30 failed) never throw. Status `failed` only occurs on uncaught exceptions. The UI shows a green badge.

7. **Messenger retry is ineffective** — Handler sets status to `failed` and flushes BEFORE rethrowing. On retry, the handler finds status ≠ `queued` and skips. Net: no automatic recovery.

8. **No concurrent-run guard for admins** — Admins bypass cooldown entirely. Rapid clicks create many parallel runs that can overwrite each other's results.

---

## Steps to reproduce

1. Have a working setup with memories stored using a 1024-dim model (e.g. Ollama bge-m3)
2. Go to AI Models config → change VECTORIZE to `text-embedding-3-large` (1536 dim)
3. Confirm the switch in the modal
4. Observe: re-vectorize run shows "completed" with "22/-8" (22 = documents/synapse only, 8 = synapse failures from different provider)
5. Send a chat message that should create a new memory (e.g. "Ich wohne in Berlin")
6. Observe worker logs: `Vector dimension error: expected dim: 1024, got 1536`
7. Memory is never stored, no user-visible error

---

### Architecture decision needed

The fundamental tension: the system assumes a **fixed 1024-dim collection** but allows selecting models with **arbitrary dimensions** in the UI.

**Option A (simpler):** Keep 1024 fixed. Filter the VECTORIZE dropdown to only show models with `meta.dimensions === 1024`. Add dimension validation in `UserMemoryService` as safety net. Fix `scrollMemories(0)` → scroll all users.

**Option B (flexible):** On model switch, recreate `user_memories` and `user_documents` collections with the new dimension (like synapse already does). Update `UserMemoryService` to read collection dimension dynamically. More work but removes the restriction.

### Additional issues in the same area

- `EmbeddingReindexService::normalizeVectorDimension()` truncates larger vectors to 1024 — this technically "works" for reindex but destroys semantic quality
- `EmbeddingMetadataService::isStale()` uses global VECTORIZE default, not per-user override — can produce false stale flags
- `chunks_total` in `RevectorizeRun` is always the sum of ALL scopes even when a specific scope is requested — progress denominator is wrong
- Documents/memories "skipped — no model" early-returns without incrementing `chunks_failed`
- Scroll failure in `reindexMemories` returns without failure accounting — run still "completed"
- `costActualUsd` entity field is never written (always 0)

### Related files

- `backend/src/Service/UserMemoryService.php` — `storeInQdrant()` (no dim check)
- `backend/src/Service/Embedding/EmbeddingReindexService.php` — `reindexMemories()` (`scrollMemories(0)`)
- `backend/src/Service/VectorSearch/QdrantClientDirect.php` — `DEFAULT_VECTOR_DIM = 1024`, no `recreateMemoriesCollection`
- `backend/src/Controller/AdminEmbeddingController.php` — `switch()` (no key check, no rollback)
- `backend/src/MessageHandler/ReVectorizeMessageHandler.php` — status logic
- `backend/src/Service/Embedding/EmbeddingModelChangeGuard.php` — admin bypass

---

## Screenshots/Logs

Worker log on dimension mismatch:
```
{"message":"Failed to upsert memory to Qdrant","context":{"point_id":"mem_1_1779096298538971","error":"Qdrant request failed [POST /collections/user_memories/points/batch?wait=true]: HTTP 400 — {\"status\":{\"error\":\"Wrong input: Vector dimension error: expected dim: 1024, got 1536\"},\"time\":0.007015452}"},"level":400,"level_name":"ERROR","channel":"app"}
```

Run history showing "completed" with 0 processed / 30 failed (Cloudflare without key):
```
BID=3  scope=all  from=88  to=188  status=completed  processed=0  failed=30
```

### LINKED ISSUE #952: Routing: MessageClassifier mapTopicToIntent missing granular Synapse-v2 topics → media generation always falls back to chat
## Problem

When Synapse Routing is disabled (the default), the AI sorter (`MessageSorter`) correctly classifies image/video/audio generation requests as `image-generation`, `video-generation`, or `audio-generation` — but `MessageClassifier::mapTopicToIntent()` does not map these topics and falls back to `chat`. All media generation requests are routed to `ChatHandler` instead of `MediaGenerationHandler`.

---

## Expected

"erstelle ein bild einer katze" → topic `image-generation` → intent `image_generation` → `MediaGenerationHandler` → actual image generated.

## Actual

"erstelle ein bild einer katze" → topic `image-generation` → intent `chat` (fallback) → `ChatHandler` → text model hallucinates a markdown image tag, no real image generated.

Confirmed by logs:

```
MessageSorter: ✅ Classification result: topic: "image-generation", language: "de"
InferenceRouter: Routing to handler: intent: "chat", topic: "image-generation"
InferenceRouter: Handler resolved: handler_name: "chat"
```

---

## Steps to reproduce

1. Ensure Synapse Routing is **disabled** (default — no `SYNAPSE_ROUTING_ENABLED` config entry)
2. Ensure a TEXT2PIC model is configured and active (e.g. `gemini-3.1-flash-image-preview`)
3. Send "erstelle ein bild einer katze" (or "generiere ein bild", "generate an image of a cat")
4. Observe: response is text-only from ChatHandler, no image generated
5. Confirm in DB: `BFILE=0`, `BFILEPATH=''` on the outgoing message

---

## Notes

**Root cause:** The granular Synapse-v2 topics (`image-generation`, `video-generation`, `audio-generation`) exist in `PromptCatalog` and `TopicAliasResolver` maps them back to `mediamaker`. However, `TopicAliasResolver` is only wired into `SynapseRouter`, not into the `MessageClassifier` path. When Synapse is disabled, `MessageSorter` produces these granular topics but `mapTopicToIntent()` only knows the legacy names (`mediamaker`, `text2pic`, etc.) and falls through to `'chat'`.

**Fix options (either one):**
1. Add the granular topics to `mapTopicToIntent()`: `'image-generation' => 'image_generation'`, `'video-generation' => 'image_generation'`, `'audio-generation' => 'image_generation'`
2. Wire `TopicAliasResolver::resolve()` into the `MessageClassifier` path before `mapTopicToIntent()` is called

**Secondary issue:** `generiere` (German imperative of "generieren") is missing from the `$mediaTriggers` fast-path bypass list (line 517). When the fast-path is enabled it would classify "generiere ein bild..." as `general` and never reach the AI sorter at all. However, this is moot for the current reproduction because `erstelle` IS in the list and the fast-path correctly defers to the sorter — the sorter classifies correctly but the intent mapping is broken.

**Affects all channels** (web, widget, email, WhatsApp) when Synapse Routing is disabled.

---

## Screenshots/Logs

N/A — log output included above.

cristiangrxs:
## Summary
Five independent `prio:1` fixes stacked on `github-issues-fixing-v6`: graceful fallback when AI structuring is unavailable for memories (#439), persisting generated media on email/widget/whatsapp outgoing messages so the web chat history renders the player (#626), gating prompt `aiModel` saves through the embedding premium guard so the per-prompt path can no longer bypass `/config/ai-models` restrictions (#891), resolving the granular Synapse-v2 topics in `MessageClassifier` so media-generation requests stop falling back to chat (#952), and making the SafeModelChange embedding switch actually safe — dimension safety net, real memories reindex, provider gate, concurrent-run guard, and BCONFIG rollback on failure (#948).

### Closes #439 — Memories: "Create with AI" graceful fallback when AI structuring is unavailable
- `UserMemoryController::parseMemory` now distinguishes AI-unavailable vs. storage-unavailable with stable `code` fields (`ai_parser_unavailable`, `ai_no_chat_model`, `memory_storage_unavailable`, `ai_parse_failed`, `input_required`), and pre-checks the resolved chat model so we fail fast with an actionable code instead of bubbling up a provider error.
- `MemoryFormDialog.vue` inspects `ApiError.details.code` and:
  - **Create + AI down** → persists the raw input as a single-action `create` with a derived key + sensible category (user choice → custom → `notes`) so the user never loses what they typed; proceeds to the preview step.
  - **Edit + AI down** → switches to advanced mode with the current memory values pre-filled.
  - **Storage down (Qdrant)** → surfaces a clear hard error (no false fallback).
- Per-case translations added in `de`/`en`/`es`/`tr`: `aiUnavailableCreateFallback`, `aiUnavailableCreateReason`, `aiUnavailableEditFallback`, `storageUnavailable`, `parseFailed`.

### Closes #626 — Mail/Widget/WhatsApp: Generated media now persisted on outgoing messages
- `WebhookController` (email + WhatsApp), `WidgetPublicController`, and `WhatsAppService` now mirror the `StreamController` behaviour: read `metadata.file.{path,type}` from the `MediaGenerationHandler` result and persist it on the outgoing `Message` row so `ChatController::getChatMessages` surfaces the `file: {path, type}` block and the web UI renders the player.
- The normalisation logic lives in a new shared `App\Service\Media\GeneratedFileMetadataNormalizer` service (extracted from the controllers as the DRY follow-up to FExB17's review on this PR). The service refuses to flip `file=1` when the handler returned an empty path, so we never store a broken file reference.
- Dedicated unit coverage on the normaliser service (replaces the earlier private-method test).

### Closes #891 — Models: prompt `aiModel` now subject to the same premium gate as `/config/ai-models`
- New `App\Service\Model\PromptModelEligibilityValidator` — single point of truth for "may this user pin model X as a prompt `aiModel`?". Unknown/inactive models → 400; VECTORIZE/EMBEDDING-tagged models go through the existing `EmbeddingModelChangeGuard::assertCanChange()` so the subscription rule from PR #908 applies to both surfaces. Future per-model premium flags belong here, not in two controllers.
- Wired into `PromptController::create` / `::update` and `LegacyApiController::promptUpdate` so every prompt-write path is gated consistently. The 403 body matches `ConfigController::saveDefaultModels` (`error: 'requires_premium'`, `currentLevel`, `message`).
- `TaskPromptsConfiguration.vue` reuses the `ApiError` class from PR #908 to surface the structured reason in the toast, matching the AIModelsConfiguration UX. Adds `config.taskPrompts.saveErrorPremiumRequired` i18n key in `en` + `de`.
- 13-case unit test on the validator (no-override sentinels `-1` / `0` / `null` / non-numeric, unknown / inactive models, premium gate for embedding tags, premium pass-through, case-insensitive tag matching).

### Closes #952 — Routing: `MessageClassifier::mapTopicToIntent` now resolves granular Synapse-v2 topics
- With Synapse Routing disabled (the default), `MessageSorter` directly emits the granular topics from `PromptCatalog` (`image-generation`, `video-generation`, `audio-generation`). `mapTopicToIntent()` only knew the legacy canonical topics, so every media-generation request fell back to `chat` → `ChatHandler` and the text model hallucinated a markdown image tag instead of routing to `MediaGenerationHandler`.
- Inject `TopicAliasResolver` into `MessageClassifier` and resolve the canonical topic + implied media type before intent mapping. `SynapseRouter` already runs the resolver internally; this mirrors the contract for the AI-sorter path. The granular topic is preserved in the result as `granular_topic` for diagnostics. `media_type` from the sorter wins over the alias-inferred value when both are present.
- Secondary fast-path miss from the same issue: the German imperatives `generiere` / `generiert` / `generier` were absent from `$mediaTriggers`, so with the default-on fast-path "generiere ein bild..." bypassed the AI sorter and was classified as `general`/chat. Added to the trigger list.
- 4 new regression cases in `MessageClassifierTest` (DataProvider × 3 granular topics, sorter-`BMEDIA`-wins, canonical-passthrough, DataProvider × 3 German imperatives).

### Closes #948 — Embedding: SafeModelChange is now actually safe (no more silent data loss)
Headline `prio:1` from FExB17: switching the VECTORIZE model used to fail open against Qdrant (HTTP 400 dim mismatch → memory silently lost), the re-vectorize job never re-embedded user memories, runs with 100% failures showed a green "completed" badge, and there was no rollback. Lands the Option-A fix from the issue body — keep the collection dimension fixed but add the missing safety nets so the switch flow either succeeds end-to-end or rolls back cleanly.

- **Live-path dimension safety net** (`UserMemoryService::storeInQdrant`): compare embedding length against `EmbeddingMetadataService::getCurrentVectorDim()` BEFORE upsert and throw a clear `RuntimeException` on mismatch. The existing handler maps that to a 503 for HTTP callers and a logged error for the worker — instead of a silent Qdrant 400 + lost memory.
- **Memories reindex no-op fix** (`EmbeddingReindexService::reindexMemories`): `scrollMemories(0, ...)` filtered on `user_id === 0` and therefore never returned a real memory. New dedicated `QdrantClientInterface::scrollAllMemoriesForReindex()` omits the user-id filter (only `active = true`) so every user's memories actually get re-embedded. Deliberately a separate method — privacy-relevant "all users" mode must not be reachable by callers that forget to pass a user id.
- **Failure semantics** (`ReVectorizeMessageHandler`):
  - Zero-success outcome (`chunks_processed === 0 && chunks_failed > 0`) → mark run as `failed` with a descriptive error and roll back the BCONFIG binding to `model_from_id`.
  - Uncaught exception → already-failed status now also triggers the rollback before the exception escapes.
  - "Documents/memories skipped — no model" early-returns now increment `chunks_failed` so the zero-success guard catches them.
  - Partial success (e.g. 9990/10) stays `completed` — rolling back would silently degrade a corpus that's mostly re-embedded against the new model.
- **Provider availability gate** (`AdminEmbeddingController`): pre-flight the target model's provider via `ProviderRegistry::getEmbeddingProvider($name)->isAvailable()`. Refuse the switch with 400 `provider_not_configured` when the API key is missing — instead of letting BCONFIG end up stuck on a broken model after a 0/N "completed" run.
- **Concurrent-run guard** (both switch endpoints): reject when `RevectorizeRunRepository::findActive()` is non-null with 409 `run_in_progress`. Closes the admin-cooldown bypass that let rapid clicks dispatch overlapping reindex jobs whose results overwrote each other.
- `App\Service\Embedding\VectorizeBindingService` extracted from the controller so the rollback path in the message handler reuses the exact same BCONFIG-upsert + `EmbeddingMetadataService::invalidate()` semantics.
- 12 new unit cases across `ReVectorizeMessageHandlerTest` (9 — happy / partial / zero-success / synapse-rollback / no-fromId / throw-then-rollback / rollback-error swallowed / re-delivery idempotency / missing run), `AdminEmbeddingControllerSynapseTest` (2 — provider-not-configured, run-in-progress), and `UserMemoryServiceTest` (1 — dimension mismatch rejected before upsert).

## Verification
- [ ] Not tested (explain)
- [x] Manual
- [x] Tests added/updated

Local pre-commit gate (mirrors `.github/workflows/ci.yml`):

| Step | Result |
|---|---|
| `make -C backend lint` | ✓ clean |
| `make -C backend phpstan` | ✓ no errors |
| `make -C backend test` | ✓ 1541 tests / 5476 assertions |
| `make -C frontend lint` | ✓ clean |
| `docker compose exec -T frontend npm run check:types` | ✓ clean |
| `make -C frontend test` | ✓ 429 tests / 51 files |
| `make -C frontend generate-schemas` | ✓ no schema diff (only 4xx annotations changed) |

## Notes
Per-issue closing references (each tracked separately so GitHub auto-closes them on merge):

- Closes #439
- Closes #626
- Closes #891
- Closes #952
- Closes #948

All five issues are `prio:1`. The prompt-model gate (#891) builds on the cross-cutting `ApiError` / toast plumbing shipped earlier in PR #908 — that PR already addressed the related toast UX, and this PR only reuses its mechanism on the prompts surface. The maintainer comment on #891 noted there is no per-model premium attribute today; the new `PromptModelEligibilityValidator` is the deliberate home for that future flag — extending the embedding guard reuse is the minimal correct fix for the actual bypass.

#948 deliberately ships the simpler Option-A from the issue body (keep collection dim fixed + safety nets). The Option-B path (recreate `user_memories` / `user_documents` collections with the new dim on every switch) is a separate, riskier refactor; the secondary cleanup items the author called out (`costActualUsd` never written, `chunks_total` per-scope, `normalizeVectorDimension()` semantic loss, per-user stale-detection) are deferred to follow-up PRs so this batch stays focused on the data-loss and broken-rollback path.

## Screenshots/Logs
N/A — backend hardening / UX fixes; the user-visible deltas are the toast text on the prompts page (matches the existing one on the models config page), the web-chat history now rendering the generated player for messages that originated via email/WhatsApp/widget, "erstelle ein bild einer katze" actually producing an image instead of a hallucinated markdown tag, and the embedding model switch flow either succeeding cleanly or rolling back to the previous model on failure.


backend/src/Controller/WidgetPublicController.php
```diff
@@ -1055,6 +1061,34 @@ private function processWidgetFile($uploadedFile, $widget, $widgetSession): arra
...
+            return null;
+        }
+
+        $path = isset($fileMeta['path']) && is_string($fileMeta['path']) ? trim($fileMeta['path']) : '';
+        $type = isset($fileMeta['type']) && is_string($fileMeta['type']) ? trim($fileMeta['type']) : '';
+
+        if ('' === $path) {
+            return null;
+        }
+
```
FExB17: Nit: `normalizeGeneratedFileMetadata()` is duplicated verbatim as a `private` method here and in `WebhookController`. Consider extracting into a shared trait or static utility to avoid copy-paste drift.

orgaralf (approved):
Besides the double named method, it looks ok... please address the method naming problem!

backend/src/Controller/WidgetPublicController.php
cristiangrxs: Addressed in 35f61db63 — extracted into `App\Service\Media\GeneratedFileMetadataNormalizer` (`final readonly` service) and injected via constructor DI into both `WebhookController` and `WidgetPublicController`. The unit test moved out of `tests/Unit/Controller/` into `tests/Unit/Service/Media/` and now exercises the service directly (no more `ReflectionMethod` hack). Thanks for the catch — it would have rotted the moment a third channel handler copied the same logic.

FExB17:
## Issue #891 — the actual root cause is not addressed

The `PromptModelEligibilityValidator` only gates `VECTORIZE`/`EMBEDDING`-tagged models. This **does not** fix the inconsistency described in #891.

### The real bug

**Config page** (`/config/ai-models`): When a `NEW` user changes *any* model (e.g. CHAT), `saveConfiguration()` in `AIModelsConfiguration.vue` sends **all** non-null defaults — including the **unchanged** VECTORIZE model ID — to `ConfigController::saveDefaultModels`. The backend runs the VECTORIZE premium gate whenever `VECTORIZE` is present in the payload and returns **403 for the entire request**, blocking CHAT/TEXT2PIC/etc. saves as collateral damage.

**Prompts page** (`/config/task-prompts`): The validator lets every non-VECTORIZE model through. A `NEW` user can freely assign any CHAT model to a prompt.

**Result:** A `NEW` user cannot change their CHAT model on the Config page (accidentally blocked by VECTORIZE in payload), but CAN change it on the Prompts page → inconsistency persists.

### Root cause chain

1. `AIModelsConfiguration.vue` → `saveConfiguration()` → builds `defaults` from **all** non-null entries in `defaultConfig`
2. Seeded defaults include VECTORIZE → VECTORIZE is always sent, even when unchanged
3. `ConfigController::saveDefaultModels` → `if (isset($data['defaults']['VECTORIZE']))` → premium gate fires → 403 → **no capability is saved**

### Fix options

**Option A (backend — recommended):** In `ConfigController::saveDefaultModels`, compare the submitted VECTORIZE model ID against the stored value. Only run the premium gate when the user is *actually changing* it:

```php
if (isset($data['defaults']['VECTORIZE'])) {
    $current = $this->configRepository->findOneBy([...]);
    if (!$current || (int) $current->getValue() !== (int) $data['defaults']['VECTORIZE']) {
        $this->embeddingChangeGuard->assertCanChange($user);
    }
}
```

**Option B (frontend):** Strip unchanged capabilities from the `defaults` payload before sending.

Either way, the Prompt page validator is a nice hardening layer for VECTORIZE/EMBEDDING, but it doesn't address the actual inconsistency reported in #891.

cristiangrxs:
@FExB17 you were right — the validator was only a hardening layer, not a fix for the actual #891 inconsistency. Addressed the real root cause in 9ae8d52ef (Option A from your writeup).

**What changed in `ConfigController::saveDefaultModels`:**

- Resolve the current global VECTORIZE binding from `ownerId=0` (same row the frontend dropdown reads via `getDefaultModels()`) and compare against the incoming id.
- Only run `EmbeddingModelChangeGuard::assertCanChange()` when the value is *actually* changing.
- Skip the per-user write AND the `EmbeddingMetadataService::invalidate()` call for an unchanged VECTORIZE echo — those are no-ops that should stay no-ops.
- Initial-bind edge case preserved: any non-zero incoming id on a fresh install still counts as a change, so the gate fires for the first switch.

**Net effect for the user-facing bug:** A NEW user can now save CHAT/TEXT2PIC/etc. on `/config/ai-models` even though the frontend keeps echoing the seeded VECTORIZE id — exactly because that echo is byte-equal to the stored value and is now silently dropped instead of tripping a 403. The previously-added `PromptModelEligibilityValidator` stays in place as the hardening layer it was always meant to be.

**Test coverage** (`tests/Unit/Controller/ConfigControllerSaveDefaultModelsTest.php`):

1. Unchanged VECTORIZE echo from a non-premium user → gate NOT called, CHAT persists, cache NOT invalidated.
2. Changed VECTORIZE from non-premium user → 403 `requires_premium`, no persist, no flush, no cache invalidation.
3. Changed VECTORIZE from eligible user → gate passes, VECTORIZE persists, cache invalidated once.
4. Initial bind (no existing global row) → any non-zero id fires the gate (defence against bypass-on-fresh-install).
5. Unauthenticated → 401.
6. Malformed payload → 400.

Thanks for the careful read — the issue would have shipped half-fixed otherwise.

PR #964, opened 2026-05-20
test: isolate vite 8.0.13 bump for E2E regression

FExB17:
Disposable test PR — isolates vite 8.0.11→8.0.13 to determine if it causes the E2E login failures seen in renovate/frontend-core (#942).

No other dependency changes. Will be closed after CI results.


FExB17:
CONFIRMED: vite 8.0.13 causes E2E login failures. This is the root cause of #942 failures.

PR #966, opened 2026-05-20
fix(e2e): robust login helper + vite 8.0.13 diagnosis

FExB17:
## Why

The E2E login helper blindly waits for a selector after clicking submit. When it fails, the only output is "Login failed: current URL is /login" — no way to tell whether the API call, the navigation, or the rendering broke.

This is why the vite 8.0.13 regression has been impossible to diagnose from CI logs alone.

## Changes

**E2E login helper** (`tests/e2e/helpers/auth.ts`):
- Captures the `/api/v1/auth/login` API response alongside the click (fail-fast on non-2xx)
- Waits for URL to leave `/login` (detects auth-state or router issues)
- Waits for the chat input selector (detects render issues)
- Each step throws a distinct, diagnostic error message

**vite 8.0.11→8.0.13** included intentionally to trigger the known regression and see which step the improved helper identifies as broken.

## Test plan

- [ ] If E2E fails: the new error messages will show exactly where (API / navigation / render)
- [ ] Once root cause is identified: follow-up fix or upstream issue

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -54,7 +54,7 @@
         "tailwindcss": "4.3.0",
         "terser": "5.47.1",
         "typescript": "6.0.3",
-        "vite": "8.0.11",
+        "vite": "8.0.13",
         "vitest": "4.1.6",
         "vue-tsc": "3.2.8"
       }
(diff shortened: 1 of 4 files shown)


FExB17:
Root cause found: Rolldown 1.0.x minification bug (TypeError: e is not a function). Closing in favor of a clean PR.

Issue #985, opened 2026-05-22
Mail Handler: Email body extraction fails for standard multipart emails and corrupts 8BIT-encoded content


FExB17:
## Problem

`InboundEmailHandlerService::extractEmailBody()` fails to extract text content from standard multipart emails and actively corrupts 8BIT-encoded bodies, causing the AI routing to receive garbled/raw MIME content instead of readable email text.

---

## Expected

The mail handler correctly extracts the plain-text (or stripped HTML) body from any standard email — regardless of MIME structure depth, transfer encoding, or character set — and passes clean, readable text to the AI for department routing.

---

## Actual

Multiple related failures in `backend/src/Service/InboundEmailHandlerService.php` on `main`:

### 1. Shallow multipart parsing — no recursion into nested parts (lines 192–231)

`extractEmailBody()` only iterates top-level `$structure->parts` and never recurses into sub-parts. Simple `multipart/alternative` emails (text/plain + text/html, no attachments) work because `text/plain` sits directly at the top level. But any email with attachments, inline images, signatures, or forwarded content wraps the body in a nested structure:

```
multipart/mixed
├── multipart/alternative       ← type=1 (multipart), NOT text/plain
│   ├── text/plain
│   └── text/html
└── application/pdf             ← attachment
```

The code checks `getMimeType($part)` against `text/plain` and `text/html`, but at the top level it only finds `multipart/alternative` (type=multipart) and `application/pdf` — neither matches. Both `$textPlain` and `$textHtml` stay empty. The fallback `imap_body()` returns the raw MIME content including boundaries, encoded parts, and binary attachment data.

This affects the majority of business emails (corporate signatures with inline images, emails with attachments, forwarded messages, etc.).

Result: AI receives raw MIME like `------=_Part_123\r\nContent-Type: text/plain; charset=utf-8\r\nContent-Transfer-Encoding: base64\r\n\r\nSGVsbG8gV29ybGQ=\r\n------=_Part_123--` instead of `Hello World`.

### 2. `imap_8bit()` / `imap_binary()` corrupt email bodies (lines 252–253)

```php
return match ($encoding) {
    1 => imap_8bit($body),      // 8BIT — WRONG: encodes TO quoted-printable
    2 => imap_binary($body),    // BINARY — WRONG: encodes TO base64
    3 => base64_decode($body),  // BASE64 — correct
    4 => quoted_printable_decode($body), // QP — correct
    default => $body,
};
```

[...]

Issue #984, opened 2026-05-22
Memories: persist in MariaDB as source of truth (data loss + GDPR)


FExB17:
## Summary

User memories are stored exclusively in Qdrant with no SQL backup. If the Qdrant collection is lost, recreated, or corrupted during a re-vectorize operation, all memories are permanently deleted with no way to recover them. This also creates GDPR compliance gaps.

---

## Problem / Motivation

- **Data loss during model switch:** A re-vectorize run that fails mid-flight (e.g. dimension mismatch #959, network error, worker crash) can leave the memories collection empty. The `recreateMemoriesCollection()` call drops the old collection before new embeddings are written — if the subsequent upserts fail, all data is gone.
- **Observed in production (dev):** Switching from bge-m3 (1024-dim) to text-embedding-3-small (1536-dim) caused Run 16 to fail on dimension mismatch, and Run 17's scroll found 0 points — all 14 user memories were permanently lost.
- **No recovery path:** Unlike RAG documents (stored in `BRAG` table with text content, re-embeddable at any time), memories exist only as Qdrant points. There is no `user_memories` SQL table.
- **GDPR compliance gaps:**
  - **Art. 17 (Right to erasure):** Deleting all personal data for a user requires scanning Qdrant by user_id filter — fragile and not auditable.
  - **Art. 20 (Data portability):** Exporting all memories for a user is difficult without a structured SQL source.
  - **Art. 15 (Right of access):** Listing all stored personal data requires Qdrant scroll queries rather than a simple SQL query.
  - **Audit trail:** No record of when memories were created, modified, or by what source (chat extraction vs. manual UI entry).

---

## Goal

Memories should be stored in a MariaDB table as the authoritative source of truth. Qdrant remains the vector index for similarity search, but all memory content, metadata, and lifecycle events are persisted in SQL. A re-vectorize or Qdrant failure can always be recovered from the SQL data.

---

## Acceptance criteria

- [ ] New `user_memories` (or similar) Doctrine entity and migration with columns: id, user_id, category, key, value, source (chat_extraction / manual / api), namespace, active, created_at, updated_at
- [ ] `UserMemoryService::createMemory()` and `updateMemory()` write to SQL first, then upsert the embedding in Qdrant
- [ ] `UserMemoryService::deleteMemory()` deletes from both SQL and Qdrant
[...]

PR #992, opened 2026-05-22
fix(memories): pin embedding model to memories collection so post-switch writes work

cristiangrxs:
## Summary
Closes the gap left open by #988: after a VECTORIZE model switch with a different dimension, new memories couldn't be saved (dim mismatch) and existing ones couldn't be searched (stale-filter dropped everything). This PR decouples the memory layer's embedding model from VECTORIZE via a *sticky* pointer so memories keep working even when VECTORIZE moves to a different-dim model.

## Changes
- **New** `App\Service\Memory\MemoryEmbeddingModelResolver` — resolves the memory-pinned embedding model with three strategies:
  1. Stored sticky pointer (`BCONFIG owner=0 group=MEMORIES setting=EMBEDDING_MODEL_ID`)
  2. Inference from the most recent memory payload's `embedding_model_id`
  3. Fallback to the active VECTORIZE default (fresh install path)
  Self-heals if the operator manually drops/recreates the collection.
- **`UserMemoryService`** — `storeInQdrant()`, `searchRelevantMemories()` and `searchMemoriesByVector()` now embed/filter against the resolver, NOT against VECTORIZE. New public helper `embedQueryForMemorySearch()` for the chat pipeline.
- **`ChatHandler`** — gains a `createMemoryVectorResolver()` companion to the shared-vector resolver. When the memory model equals VECTORIZE (typical case) it reuses the shared vector (no perf regression); when they diverge it embeds separately so reads land in the right vector space.
- **`EmbeddingReindexService::reindexMemories()`** — advances the sticky pointer to the new model only after a successful re-index. The pointer stays put on probe-dim aborts so writes continue to land in the still-intact old collection.
- **i18n** — `EmbeddingSwitchModal` banner reworded: replaces the misleading *"new memories continue to be stored with the active model"* (technically false after a dim-changing switch) with an accurate *"memories stay on their own embedding model"*.

## Verification
- [x] Tests added/updated
- [x] Manual (PHPStan + lint + full backend suite + vue-tsc + vitest)

New tests:
- `MemoryEmbeddingModelResolverTest` — covers all four resolution branches plus the self-heal-on-drift path (5 tests).
- `UserMemoryServiceTest::testCreateMemoryEmbedsAgainstStickyMemoryModelNotVectorize` — pins the regression so any future caller that reaches for VECTORIZE again is rejected.
- `EmbeddingReindexServiceTest` — sticky pointer advances on success, stays put on abort.

Full pre-commit gate green:
- ``make -C backend lint`` — 0 fixes
- ``make -C backend phpstan`` — ``[OK] No errors``
- ``make -C backend test`` — 1659 tests, the only error is the pre-existing Mailhog DNS issue on the integration ``MailerTest``
- ``make -C frontend lint`` — clean
- ``npm run check:types`` (vue-tsc) — clean
- ``make -C frontend test`` — 549 / 549 green

## Notes
Refs #985. Built on top of `fix/embedding-switch-memory-data-loss` (#988) — base branch is set accordingly so this PR will collapse into #988 once that lands, or against ``main`` if #988 squashes first.

The architectural choice (sticky pointer in BCONFIG instead of e.g. a separate collection per model) keeps the migration scope tiny: no schema change, no Qdrant reshuffling. A future *explicit* memories re-index (still admin-only per the team consensus from issue #984) advances the pointer cleanly via `MemoryEmbeddingModelResolver::rememberModel()`.

## Screenshots/Logs
n/a — backend / pipeline change. The frontend banner i18n change is visible to admins on the embedding-switch modal but does not change layout.

### DIFF
--- a/backend/src/Service/Embedding/EmbeddingReindexService.php
+++ b/backend/src/Service/Embedding/EmbeddingReindexService.php
@@ -7,6 +7,7 @@ namespace App\Service\Embedding;
 use App\AI\Service\AiFacade;
 use App\Entity\RevectorizeRun;
 use App\Repository\RevectorizeRunRepository;
+use App\Service\Memory\MemoryEmbeddingModelResolver;
 use App\Service\Message\SynapseIndexer;
 use App\Service\VectorSearch\QdrantClientInterface;
 use Doctrine\DBAL\Connection;
@@ -51,6 +52,7 @@ final readonly class EmbeddingReindexService
         private AiFacade $aiFacade,
         private SynapseIndexer $synapseIndexer,
         private EmbeddingMetadataService $embeddingMetadata,
+        private MemoryEmbeddingModelResolver $memoryEmbeddingResolver,
         private RevectorizeRunRepository $runRepository,
         private Connection $connection,
         private LoggerInterface $logger,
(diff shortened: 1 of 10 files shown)


cristiangrxs:
Closing — folding this commit directly into #988 to avoid stacked-branch noise. The same commit (03d4c0738) is being fast-forwarded onto fix/embedding-switch-memory-data-loss.

Issue #1000, opened 2026-05-25
Chat: Fast-path classifier cannot detect web search intent


FExB17:
## Problem

The fast-path classifier in `MessageClassifier` always returns `web_search: false` and has no ability to detect web search intent on its own. Messages that need live data but don't match the narrow `$searchTriggers` blocklist are misclassified as plain chat — the AI never performs a web search.

---

## Expected

The fast-path should detect web search intent using a cheap keyword heuristic (like `SynapseRouter::detectWebSearchIntent()` already does in <1ms), rather than always returning `web_search: false` or requiring an expensive AI sorter fallback.

## Actual

The fast-path has only two modes:
1. Return `web_search: false` immediately (broken for search-relevant queries)
2. Fall through to the AI sorter via a `$searchTriggers` blocklist (expensive, incomplete)

Adding more keywords to `$searchTriggers` is whack-a-mole — it forces the slow AI sorter path for an ever-growing list of terms. The real fix is to give the fast-path its own web search detection using the same keyword+year-pattern heuristic that `SynapseRouter` already uses.

Additionally, `BRAVE_SEARCH_ENABLED=false` in `backend/.env` silently disables the entire feature even when the API key is configured.

---

## Steps to reproduce

1. Send "günstigsten tankstellen in 10km umgebung von 48161" (no `$searchTriggers` match)
2. Fast-path classifies as `web_search: false`, skips AI sorter
3. AI hallucinates "Ich starte die Suche..." but Brave Search is never called
4. Compare with "aktuelle spritpreise in münster" which happens to reach the AI sorter and works

---

## Notes

- Core architectural issue: fast-path sets `web_search: false` unconditionally (line 96 of MessageClassifier)
- Fix: reuse `SynapseRouter`'s keyword+year-pattern heuristic inside the fast-path return, remove `$searchTriggers` blocklist entirely
- `BRAVE_SEARCH_ENABLED` should default to `true` when API key is set (or at minimum not silently no-op)
- related: #974

---

## Screenshots/Logs
// Fast-path wrongly skipped the sorter: "classification":{"topic":"general","language":"en","web_search":false,"source":"fast_path_heuristic","skip_sorting":true}

Issue #974, opened 2026-05-21
Web Search: Fast-path classifier misses German cost/price queries — no search triggered


FExB17:
## Problem

German queries like "Was kostet ein Flug nach Bergen?" or "Was kostet ein Kebap-Gericht in Münster?" never trigger web search — neither via WhatsApp nor the web chat browser. No sources are shown, the AI hallucinates "Ich starte eine Suche" but never delivers results.

---

## Expected

Messages asking for current prices/costs should trigger the Brave Search pipeline so the AI answers with real data and surfaces source citations.

## Actual

1. `MessageClassifier::canFastPathClassify()` returns `true` — message doesn't match any `searchTriggers`
2. Fast-path returns `web_search: false`, `source: 'fast_path_heuristic'`
3. `SynapseRouter::detectWebSearchIntent()` (which has better keywords) is **never called**
4. AI model is invoked without search results → hallucinates "Ich starte eine Suche" but never delivers
5. No sources/citations are shown in either WhatsApp or web chat

---

## Steps to reproduce

1. Send "Was kostet ein Flug nach Bergen?" via WhatsApp or web chat (without manually toggling web search)
2. Observe: AI responds saying it will search but no actual results or citations appear
3. Same with "Was kostet ein Kebap-Gericht in Münster?" — no search triggered

Tested on: WhatsApp (phone) and web chat (PC browser) — same behavior on both.

---

## Notes

**Root cause is in two places:**

### 1. `MessageClassifier::canFastPathClassify()` (primary)

`backend/src/Service/Message/MessageClassifier.php` line 569–574:

The `searchTriggers` list that prevents fast-path is too narrow for German:
```
'search the web', 'google ', 'latest news', 'current price',
'today\'s', "today's", 'right now', 'this week',
'aktuelle nachrichten', 'durchsuche das internet',
'busca en', 'recherche dans'
```

Missing patterns: `'was kostet'`, `'wie teuer'`, `'kost'` (stem), `'preis'`, `'flug nach'`, `'flüge'`

### 2. `SynapseRouter::WEB_SEARCH_KEYWORDS` (secondary, never reached)

`backend/src/Service/Message/SynapseRouter.php` line 151–158:

Uses infinitive `'kosten'` but `str_contains('kostet', 'kosten')` = `false`. German conjugation breaks the substring match. Should use stem `'kost'`.

**Validated:**
```
str_contains('was kostet ein flug nach bergen?', 'kosten') = false
[...]

PR #1003, opened 2026-05-25
The "Internet Search" toggle on a task prompt never enabled web search

orgaralf:
## Summary

Internet search strong trigger

## Changes

The Vue config wrote `tool_internet_search` to BPROMPTMETA while every router read the canonical `tool_internet`. The streaming pipeline additionally lacked a positive trigger for the prompt flag.
- PromptService canonicalises legacy `tool_internet_search`/`tool_files_search` keys on both read and write (canonical wins on conflict).
- MessageProcessor::processStream() now mirrors process(): positive trigger on `tool_internet=true` and hard-disable on `tool_internet=false` that beats frontend flag, AI classifier and legacy source fallback.
- Consolidated "Web search decision" log + warning when Brave is disabled.
- SynapseRouter::WEB_SEARCH_KEYWORDS expanded from 15 to ~100 stems across real estate, travel, finance, politics, sports, news, reviews, weather and tech releases so search triggers more aggressively for time-sensitive queries.
- Doctrine migration renames legacy BPROMPTMETA rows in place; duplicates collapse onto the canonical row.
- Frontend writes the canonical key names going forward and reads both for backward compat during the rollout. Regression coverage: PromptServiceTest, MessageProcessorTest (positive trigger + hard disable) and 25 new data-provider cases in SynapseRouterTest.

## Verification
- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -208,7 +208,15 @@ FFMPEG_BINARY=/usr/bin/ffmpeg
 
 # --- Brave Web Search ---
 # Get API key: https://api.search.brave.com/
-BRAVE_SEARCH_ENABLED=false
+#
+# Default is `true` because the project-wide routing policy
+# (`WebSearchTopicPolicy`) treats internet search as the default for every
+# chat. Without a configured `BRAVE_SEARCH_API_KEY`, `BraveSearchService::isEnabled()`
+# still returns false, so leaving the key empty is the right way to
+# opt out — no need to also flip `BRAVE_SEARCH_ENABLED=false`. Set it
+# to `false` only if you have an API key but want to forcibly suppress
+# all search calls anyway (testing/cost-cap scenarios).
+BRAVE_SEARCH_ENABLED=true
 BRAVE_SEARCH_API_KEY=
 BRAVE_SEARCH_API_URL=https://api.search.brave.com/res/v1
 BRAVE_SEARCH_COUNT=10
--- a/backend/migrations/Version20260525220000.php
+++ b/backend/migrations/Version20260525220000.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Canonicalise tool-flag keys in BPROMPTMETA.
+ *
+ * Historic Vue config wrote `tool_internet_search` and `tool_files_search`
+ * to `BPROMPTMETA`, while the routing layer (`SynapseRouter`,
+ * `MessageSorter`, `MessageProcessor`) only reads the canonical short
+ * names `tool_internet` and `tool_files`. The user-facing "Internet
+ * Search" / "Files Search" toggles in the Task Prompts settings page
+ * therefore never enabled web search via the streaming chat pipeline
+ * (the negative gate in `MessageProcessor::processStream()` honoured
+ * `tool_internet_search`, but no positive trigger ever did).
+ *
+ * This migration renames every legacy row in place so existing installs
+ * benefit from the bug fix without depending on PromptService's
+ * read-time alias fallback. Idempotent: re-running is a no-op once the
+ * legacy keys are gone. Rows where both the legacy and the canonical
+ * key already exist for the same prompt are deduplicated by dropping
+ * the legacy row.
+ *
+ * Note: `BPROMPTMETA` has no unique key on (BPROMPTID, BMETAKEY), only
+ * separate indexes — so a naive UPDATE would risk creating duplicate
+ * canonical rows. The dedupe DELETE runs first to prevent that.
+ */
+final class Version20260525220000 extends AbstractMigration
+{
+    /**
+     * Map of legacy key → canonical key. Keep in sync with
+     * {@see \App\Service\PromptService::METADATA_KEY_ALIASES}.
+     *
+     * @var array<string, string>
+     */
+    private const ALIASES = [
+        'tool_internet_search' => 'tool_internet',
+        'tool_files_search' => 'tool_files',
+    ];
+
+    public function getDescription(): string
+    {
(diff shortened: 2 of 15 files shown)


orgaralf:
Thanks @copilot — all three are valid false-positive risks and contradict the new `nonSearchWorthyQueryProvider` contract added in the same PR.

Fixed in e70ea40:

- **`wm `/`em `** dropped → replaced with the unambiguous full stems `europameister` / `weltmeister` (both catch the `-schaft` suffix via prefix substring match) and `turnier`.
- **`war `** dropped → `krieg`/`konflikt`/`conflict`/`crisis` cover the conflict domain; English war-news queries co-occur with `today`/`current`/year patterns which already trigger.
- **`best `/`beste `** dropped → `top 10`, `vergleich` and the year pattern cover legitimate listicle/recommendation queries.

Added 9 negative regression cases (`em_collision_system`, `war_collision_einstein`, `best_collision_signoff_de`, …) that mirror your exact examples, plus 3 positive cases for the sports replacements. Trade-off kept conservative: substring architecture stays, no regex/word-boundary rewrite — short and ambiguous stems are simply not allowed in the list.

orgaralf:
**Relation to #1000 — does NOT close that issue.**

For transparency: this PR touches the per-prompt `tool_internet` toggle and the `SynapseRouter` keyword list, but leaves the root cause of #1000 untouched:

- `MessageClassifier::classify()` still returns `'web_search' => false` hardcoded on the fast-path (Z. 96).
- `canFastPathClassify()` still uses the narrow `$searchTriggers` blocklist (Z. 576-587) as the only escape hatch.
- `SynapseRouter`'s expanded keyword list does **not** apply on the fast-path — those are two independent code paths, and Synapse routing is opt-in BETA (off by default).

What this PR does help with for #1000's example query (`"günstigsten tankstellen …"`):

- If the user has the per-prompt **Internet Search** toggle ON, `MessageProcessor::processStream()`'s new `tool_internet=true` positive trigger fires *after* the fast-path returned `web_search: false` — so search is still performed.
- If the toggle is off (default), behaviour is unchanged from before this PR. The fast-path still misses time-sensitive queries that don't appear in `$searchTriggers`.

**Proper fix for #1000 (separate PR):**
1. Extract `SynapseRouter::detectWebSearchIntent()` into a shared `WebSearchIntentDetector` (or inject via DI).
2. Replace the hardcoded `'web_search' => false` in `MessageClassifier::classify()` fast-path return with a call to the shared detector.
3. Remove the `$searchTriggers` blocklist in `canFastPathClassify()` entirely — it becomes redundant once the fast-path can decide `web_search` itself.
4. Make `BraveSearchService::isEnabled()` opt-out (`enabled unless BRAVE_SEARCH_ENABLED=false`) when an API key is configured, to address the "silent no-op" half of #1000.

The `tool_internet` UI override added here stays as an intentional per-prompt control layer on top of that consolidated heuristic.

orgaralf:
**Update: pushed `238fc0a` — refactor closes #1000 and #974.**

Per [the discussion](https://github.com/metadist/synaplan/pull/1003#issuecomment-4537272846), the proper architectural fix for #1000 is now in this PR:

## What changed

The five-trigger gating chain (frontend flag → prompt opt-in → AI classifier → legacy source → prompt opt-out) is replaced by a single project-wide policy in `WebSearchTopicPolicy::shouldSearch()`:

1. Prompt has explicit `tool_internet=true` → **search** (beats every gate)
2. Topic is a NON_WEB_SEARCH topic → **no search** (asset/doc generation handlers don't consume web context)
3. Prompt has explicit `tool_internet=false` → **no search** (user opt-out)
4. Otherwise (`tool_internet=null`, never customised) → **search** (project default)

Translated: **search is the default**; the only ways to suppress it are explicit per-prompt opt-out or a media-generation topic.

## Closes #1000

The `MessageClassifier` fast-path no longer pre-empts the search decision — it returns `web_search => null` and the policy decides downstream in `MessageProcessor`. The `$searchTriggers` blocklist (the whack-a-mole that #1000 called out) is **deleted**. The `MessageClassifierTest` regression suite includes the original failing query `günstigsten tankstellen in 10km umgebung von 48161` and asserts it now stays on the fast-path AND triggers search via the policy.

Also addresses the second half of #1000 (silent no-op): `.env.example` default for `BRAVE_SEARCH_ENABLED` flipped from `false` to `true`. New installs with an API key get search out of the box; existing deployments with explicit `=false` are respected.

## Closes #974

The `kost` stem regression case from #974 is preserved in `MessageClassifierTest::germanCostQueryFastPathProvider` but the assertion changed: previously it tested "fast-path defers to AI sorter for cost queries" (workaround), now it tests "fast-path takes the query AND web_search hint stays null so the policy can fill it in" (proper fix). Same effect end-to-end — search fires — but no fragile keyword blocklist to maintain.

## Cleanup as a free side-effect

- `SynapseRouter::WEB_SEARCH_KEYWORDS` (~100 stems) — **deleted**
- `SynapseRouter::YEAR_PATTERN` — **deleted**
- `SynapseRouter::detectWebSearchIntent()` — **deleted**
- `MessageSorter` BWEBSEARCH parsing — now advisory (the LLM's vote is logged but doesn't gate)
- `PromptService::loadMetadataForPrompt()` no longer pre-populates `tool_internet => false`, which was collapsing the "no opinion" state onto "explicit opt-out" (subtle silent bug)

Net diff: **−546 / +262** lines across 11 files, despite the architectural depth — the new policy is much simpler than the multi-trigger logic it replaces.

## Verification

- `WebSearchTopicPolicyTest`: 14 cases covering all 4 rules × topic-state combinations
- `MessageProcessorTest`: 2 new integration tests (default search for chat, no search for media-gen)
- `MessageClassifierTest`: rewritten #974 regression incl. the #1000 tankstellen case
- `SynapseRouterTest`: 36 tests, all green; heuristic-driven tests replaced with policy-driven ones

Full local quality gate: `composer lint` ✅, `composer phpstan` ✅ (0 errors), `1700 tests` ✅.

PR #1006, opened 2026-05-27
fix: model update with Google Flash

orgaralf:
## Summary
Google was not working

## Changes
Check for Flash names

## Verification
- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated


FExB17 (commented):
Two real issues I spotted:

**1. BID 229 (Gemini 3.1 Flash TTS) will silently be ignored**

`GoogleProvider::synthesizeSpeech()` has a hardcoded allowlist that only permits `gemini-2.5-flash-preview-tts` and `gemini-2.5-pro-preview-tts`. Any other model ID gets overwritten back to the 2.5 default. So the new TTS catalog entry will never actually be used — it'll always fall back silently.

**2. BID 170 repurposed from `gemini-2.5-flash` → `gemini-3.5-flash`**

Existing users who have BID 170 configured in their widgets will silently get a different model. Gemini 2.5 Flash still exists at Google. Safer approach: keep BID 170 as-is and add 3.5 Flash as a new BID (the vision twin already got its own slot at 223).

---

The OpenAI `analyzeImage` fix with `usesCompletionTokens()` looks correct.

FExB17 (approved):

PR #1053, opened 2026-06-11
feat(multitask): per-task errors with retry, email_me task, simpler Again UX, WhatsApp file fixes

### LINKED ISSUE #1055: Multitask: Generated document file disappears after page reload
## Problem

In multi-node task plans, the first generated file (typically a document like DOCX) is not registered as a `File` entity in the ManyToMany relation, so it disappears from the chat UI after a page reload.

---

## Expected

After a page refresh, the generated document should still be visible as a download badge on the assistant message — identical to how it looks during streaming (the TaskPlanBubble "Document ✓ Done" with Download link).

## Actual

After refresh, the document download link is completely gone. Only the text content and search results remain. This affects both the PWA and regular browser.

---

## Steps to reproduce

1. Open a chat and send a message that triggers a multi-node task plan with document generation (e.g. "fasse das zusammen und pack es in eine Word-Datei")
2. Observe the streaming response: TaskPlanBubble shows "Answer ✓ Done" and "Document ✓ Done" with a Download link — works correctly
3. Refresh the page (F5 / Cmd+R)
4. The document download link is gone; only the translated text "Hier ist die Word-Datei..." remains without any download option

---

## Notes

**Root cause:** In `StreamController.php`, the first file from `metadata['file']` (index 0) is only persisted via the legacy `setFile(1)` + `setFilePath()` fields. It is never registered as a `File` entity via `registerExistingGeneratedFile()` + `addFile()`. Extra files (index > 0) ARE correctly registered.

The `ChatController::getMessages()` endpoint serializes only the ManyToMany `getFiles()` collection — the legacy single-file field is only used for image/video/audio rendering in the frontend, not for documents.

During streaming, the download works because `TaskPlanBubble` renders `card.url` directly from the `task_file` SSE event. But `taskPlan` state is not persisted, so it vanishes on reload.

**Non-multitask document generation is unaffected** — it goes through the `storeGeneratedFileInStream()` → `addFile()` path correctly.

**Fix:** Register the first file (index 0) from `metadata['file']` as a `File` entity and call `addFile()` on the outgoing message, same as is already done for extra files at index > 0.

Relevant code:
- `backend/src/Controller/StreamController.php` lines 1040–1058 (first file handling) vs. 1065–1083 (extra files)
- `backend/src/Service/Multitask/Execution/Runner/DocumentGenerationRunner.php` (returns `local_path` in descriptor)
- `frontend/src/components/ChatMessage.vue` line 397 (`taskPlan?.active` gate)

---

## Screenshots/Logs

N/A — reproducible on every multi-node document generation followed by a page refresh.

### LINKED ISSUE #1057: Multitask: Streamed answer text not rendered until chat switch/reload
## Problem

When a multitask (DAG) plan executes, the final assembled answer text is not rendered in the chat bubble during the live stream. Only metadata (Web Search badge, Sources carousel, TaskPlanBubble cards) is visible. The text appears after switching chats or reloading the page.

---

## Expected

The assembled answer text should be visible in the chat bubble immediately after streaming completes, without requiring a page reload or chat switch.

## Actual

After the stream finishes:
- "Web Search · N" badge is visible
- "Sources (N)" carousel is visible
- TaskPlanBubble shows task cards as "Done"
- **Answer text is missing** — `message.parts[0].content` remains empty

Switching to another chat and back (which triggers a fresh `GET /api/v1/chats/{id}/messages`) renders the full text correctly, confirming backend persistence is fine.

---

## Steps to reproduce

1. Send a message that triggers multitask routing with web search (e.g. upload an image and ask a factual question about it)
2. Observe the streaming response: TaskPlanBubble shows tasks completing, Sources appear, but no answer text is rendered
3. Switch to another chat, then switch back
4. The full answer text is now visible

---

## Notes

**Root cause (frontend):** In `ChatView.vue`, when `taskPlan.active` is `true`, all SSE events with `status === 'data'` are explicitly suppressed (~line 1892):

```typescript
} else if (
  (data.status === 'data' || data.status === 'file' || ...) &&
  historyStore.messages.find((m) => m.id === messageId)?.taskPlan?.active
) {
  // Multitask mode: suppress normal single-bubble content/media events.
}
```

The backend's `ResultAssembler` emits the final assembled text as a standard `data` chunk after DAG completion. Since `taskPlan.active` is never cleared before `complete`, this chunk is dropped. The `complete` handler then skips `renderStreamingContent()` because `fullContent` is empty.

On reload, `loadMessages()` fetches the persisted `m.text` from the API and parses it fresh — bypassing the streaming path entirely.

**Possible fixes:**
1. On `complete`: if `fullContent` is empty but `backendMessageId` is set, re-fetch the single message from the API as a safety net
2. Allow the final assembled `data` chunk through even when `taskPlan.active` (e.g. after all task cards are `done`)
3. On `complete`: set `taskPlan.active = false` before processing, so a late `data` chunk is not suppressed

related: #1055 (inverse problem — files visible during stream but lost on reload)

---

## Screenshots/Logs
before refresh:
Image

after refresh:
Image

orgaralf:
## Summary

Follow-up fixes for the multi-task (DAG) release (#1048):

- **Per-task error messages + retry**: `DagExecutor` now sends the truncated error (plus resolved prompt/media type for media nodes) with failed/skipped `task_update` events. Failed media task cards show the reason and a **Retry with <next model>** button that re-runs just that step with the next model from the catalog (round-robin). The button is hidden when the capability pool only contains the model that just failed.
- **`email_me` DAG task**: new capability so prompts like *"write a spring poem, read it aloud, make a fitting image — and mail it to me"* show the result in chat **and** send a multi-MIME mail (markdown→HTML body, first image inline, remaining files attached) to the verified account-owner address. Guarded against placeholder/unverified emails; attachment paths are resolved strictly inside the uploads dir. Subject/confirmation strings localized for all four app locales (en/de/es/tr).
- **Simpler "Again" UX**: DAG turns persist a `multitask` meta flag (also exposed through chat history), so the frontend shows a plain **Again** (full re-plan) for DAG answers and keeps **Again with <model>** only for single-prompt turns.
- **WhatsApp multi-file delivery**: response text is sent as its own message when a caption can't carry it (audio has no captions; captions cap at 1024 chars), `document` is now a first-class type for primary + extra files, and extra-file sends are best-effort with per-file error logging.
- **Ops runbook**: documented the prod `BCONFIG` update to default text2pic/pic2pic to Nano Banana 2 (BID 190, `BOWNERID = 0` only) and the `APP_URL` reachability check WhatsApp media depends on.

## Bug fixes folded in from review

- **Multitask answer text now renders without a reload** — the assembled DAG text is streamed once after execution and was dropped while `taskPlan.active`; it now accumulates so the `complete` flush renders it in the bubble (the `compose_reply` node has no card of its own). Covered by a new e2e assertion. Fixes #1057
- **Generated documents survive a page reload** — the primary multi-task output file (`metadata['file']`) is now registered as a `File` entity and attached to the OUT message whenever the legacy single-file channel can't render it on reload (documents/`.ics`); image/video/audio stay legacy-only so reload doesn't duplicate the inline media/player. Covered by `StreamControllerTaskPlanFilesTest`. Fixes #1055
- **`fix(chats)`: a late `POST /api/v1/chats` response no longer steals the active chat selection** (race between ChatView boot auto-create and a faster "New Chat" click) — this was the intermittent chromium E2E failure.

## Test plan

- [x] `make lint` (backend + frontend)
- [x] `make test` (full backend PHPUnit 1905 tests + frontend Vitest 610 tests)
- [x] Routing characterization snapshots re-recorded — no drift
- [x] E2E `multitask.spec.ts` (chromium, CI-like test stack on :8001) — task cards complete **and** the answer text renders without a reload
- [x] Manual: trigger a DAG with a failing image model → error text + "Retry with…" on the card
- [x] Manual: "… and mail it to me" prompt → multi-MIME mail with audio + inline image
- [ ] Manual: WhatsApp DAG turn with text + image + audio → all three arrive

Fixes #1055
Fixes #1057

### DIFF
--- a/AGENTS_DEV.md
+++ b/AGENTS_DEV.md
@@ -133,6 +133,69 @@ make -C backend test
 - ⚠️ **`BCONFIG` defaults are bootstrap-only.** Changing a value in `DefaultModelConfigSeeder` / `RateLimitConfigSeeder` does NOT propagate to existing installs (the `INSERT IGNORE` skips rows that already exist). To force-roll-out a new default everywhere, ship a dedicated migration that explicitly UPDATEs the affected `BCONFIG` rows.
 - ✅ On first boot against a legacy prod DB (`BUSER` exists but `doctrine_migration_versions` does not): the entrypoint registers ONLY the baseline migration. Every post-baseline migration runs normally via `doctrine:migrations:migrate`, so legacy installs receive new schema changes the same way fresh installs do.
 
+### ⚠️ Production Platform Specifics (`synaplan-platform` + external Galera cluster)
+
+Production is **NOT** the `synaplan/` docker-compose stack. It is the `synaplan-platform/`
+repo deploying the pre-built `ghcr.io/metadist/synaplan:latest` image, talking to a
+**MariaDB Galera cluster that runs OUTSIDE Docker**. These facts have caused several
+production-only migration failures that never reproduce in local dev. Internalize them.
+
+**Topology**
+- Nodes `web1` / `web2` / `web3` all point at the **same** Galera database
+  (`host.docker.internal:3306` from each node's backend container). There is **one**
+  shared schema, not three. Galera replicates DDL **synchronously**, so any manual SQL
+  fix only needs to run on **one** node.
+- Migrations run on backend container start. On the shared DB the first node to come up
+  runs them; the others then see them already applied.
+- Node IPs and the cluster inventory live in the **private** `synaplan-platform` repo —
+  never commit them here (this repo is public).
+
+**1. `DATABASE_*_URL` MUST use the `mariadb-` serverVersion prefix**
+- `synaplan-platform/.env` sets `DATABASE_WRITE_URL` / `DATABASE_READ_URL`. The
+  `serverVersion=` query param **overrides** `doctrine.yaml` per connection.
+- It **must** read `serverVersion=mariadb-<x.y.z>` (e.g. `mariadb-12.1.2`). DBAL picks
+  `MariaDBPlatform` only when the version string contains `mariadb`
+  (`AbstractMySQLDriver` does `stripos($version, 'mariadb')`). A bare `12.1.2` falls
+  through to the **MySQL** platform, whose introspection mishandles identifiers and
+  MariaDB's quoted string defaults → phantom diffs and lowercase `TableDoesNotExist`.
+- Confirm the cluster's version string with `SELECT VERSION();` (it reads like
+  `<x.y.z>-MariaDB-...`) and keep the number in sync, but the **prefix** is what selects
+  the platform.
+
+**2. Migrations that may run incrementally on prod MUST NOT touch `Schema $schema`**
+- Reading the injected schema (`$schema->hasTable()/getTable()/hasColumn()`) forces
+  doctrine/migrations' `LazySchemaDiffProvider` to introspect and run the **DBAL schema
+  comparator**. On this Galera DB the comparator throws
+  `There is no table with name "<x>" in the schema.` — a MariaDB FK identifier-resolution
+  quirk. The named table **varies per run** (`buser`, `bsessions`, `bprompts`, …) because
+  `information_schema` is returned unordered across the cluster, which is the fingerprint
+  of this bug. Pure `addSql()` migrations never materialize the diff and are unaffected.
+- ✅ Use raw, comparator-free, **idempotent** SQL instead of Schema reads:
+  - `CREATE TABLE IF NOT EXISTS ...`
+  - `ALTER TABLE x ADD COLUMN IF NOT EXISTS ...` / `DROP COLUMN IF EXISTS ...`
+  - `DROP TABLE IF EXISTS ...`
+  - For conditional DML, guard with a plain query on `$this->connection` (NOT the Schema
+    API), e.g.
+    `((int) $this->connection->fetchOne("SELECT COUNT(*) FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'X' AND COLUMN_NAME = 'Y')) > 0`.
+- ❌ Never gate prod-reachable migration DDL on `$schema->hasTable(...)` etc.
+
+**3. Foreign keys without `ON DELETE CASCADE`**
+- Several FKs (e.g. `BPROMPTMETA.BPROMPTID → BPROMPTS.BID`) have **no** cascade. A
+  migration that deletes parent rows must delete the child rows **first**, or it fails
+  with `1451 Cannot delete or update a parent row`. Local dev often has no child rows, so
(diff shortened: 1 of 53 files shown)


FExB17 (commented):
## Review Notes

### Minor: Single-model retry offers the same failed model

`TaskCard.vue` — when there's only one model in the capability pool:
```ts
if (options.length === 1) return options[0]
```
The button shows "Retry with <same model that just failed>". `retryModel` should return `null` here so `canRetry` becomes false and the button is hidden — retrying the identical model on a non-transient failure is misleading.

### Missing backend translations for ES/TR

`emails.de.yaml` and `emails.en.yaml` have the new `task_result` keys, but `emails.es.yaml` and `emails.tr.yaml` are missing them. The `EmailMeRunner` resolves the user's locale — Spanish/Turkish users get an English subject line silently.

FExB17:
Two open issues from today fall squarely within the scope of this PR and should be addressed here:

- **#1057** — Multitask: Streamed answer text not rendered until chat switch/reload. This PR already modifies the `taskPlan.active` streaming logic in `ChatView.vue` but does not fix the suppression of the final assembled `data` chunk. The `complete` handler runs while `taskPlan.active` is still `true`, so the ResultAssembler's text is dropped.

- **#1055** — Multitask: Generated document file disappears after page reload. The PR improves multi-task file delivery (WhatsApp documents, extra files) and touches `StreamController.php` for the `multitask` meta — but `metadata['file']` (index 0) is still never registered as a `File` entity via `addFile()`, so documents vanish on reload.

Both affect the exact code paths this PR modifies.

orgaralf:
All review points addressed in 15f361fd9:

- **#1057 (answer text not rendered until reload)**: the suppression branch in `ChatView.vue` now accumulates `data` chunks into `fullContent` while `taskPlan.active` (no live single-bubble rendering), so the `complete` flush renders the assembled text immediately. The `compose_reply` node is `hidden` and has no card, so nothing is duplicated. New e2e assertion in `multitask.spec.ts` verifies the text renders without a reload — green against the CI-like test stack.
- **#1055 (document vanishes after reload)**: `StreamController` now registers the primary task-plan file (index 0) as a `File` entity + `addFile()` via the new `persistTaskPlanFiles()` helper — but only when the legacy single-file channel can't render it on reload (document/other types). image/video/audio stay legacy-only, otherwise reload would show a duplicate inline media/player (`files[]` audio entries also get a player part). Unit-tested in `StreamControllerTaskPlanFilesTest`.
- **Single-model retry**: `TaskCard.retryModel` returns `null` for pools with ≤ 1 model, hiding the button instead of offering the model that just failed. Unit test added.
- **ES/TR email translations**: `emails.es.yaml` / `emails.tr.yaml` didn't exist at all (not just missing keys) — created both with the full catalog translated, including the new `task_result` keys, matching the app's supported locales.

cristiangrxs (approved):

PR #1056, opened 2026-06-11
fix(deps): ignore phpspreadsheet security advisories blocked by phppresentation

FExB17:
## Summary

- Adds documented ignore entries for 7 phpspreadsheet security advisories to `composer.json` audit config
- All CVEs affect **all versions** including v5 (<=5.6.0) — no fixed version exists upstream
- `phppresentation` requires `phpspreadsheet ^4.0` and blocks the v5 upgrade path
- Unblocks Renovate lockfile maintenance (PR #946) which has been failing due to Composer blocking the advisory-affected packages

## Context

The ignore entries include a reason string explaining why each advisory is ignored. Once upstream provides fixes, the ignores should be removed.

## Test plan

- [x] `composer validate` passes
- [x] `composer audit` exits 0 (all advisories properly ignored)
- [x] No code changes — config only

### DIFF
--- a/backend/composer.json
+++ b/backend/composer.json
@@ -72,7 +72,15 @@
             "codewithkyrian/platform-package-installer": true
         },
         "audit": {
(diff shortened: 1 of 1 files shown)


orgaralf (approved):
Thanks!

PR #1073, opened 2026-06-12
fix(chat): stop reasoning chain-of-thought leaking into visible responses

### LINKED ISSUE #1067: Chat: internal AI reasoning/chain-of-thought leaking into visible response (regression)
## Problem

The AI's internal reasoning ("internal dialogue") is displayed as part of the visible response text in production. Users see the model's planning/thought process instead of only the final answer. Additionally, **system prompt instructions and internal architecture are leaked** to end users.

---

## Expected

Only the final answer is shown to the user. Internal reasoning, system prompt instructions, and RAG pipeline details must never be visible.

---

## Actual

The full chain-of-thought is rendered as regular message text, exposing:

**1. System prompt leakage — internal instructions visible:**
> "The instruction: answer in German, produce only requested content, no preamble."
> "The instruction: translate the user text into 'en'. Return ONLY the translation."

**2. Architecture exposure — RAG/search pipeline details revealed:**
> "The user gave web search results for 'wie wird das Wetter in Münster heute'."

The user did NOT provide web search results — the system injected them. This reveals that search results are passed in the `user` role rather than as `tool`/`system` messages, causing the model to attribute them to the user.

**3. Prompt injection vector:**
If the model cannot distinguish between real user input and system-injected content (search results, RAG context), an attacker could craft inputs that override system instructions.

---

## Steps to reproduce

1. Open `web.synaplan.com` and log in
2. Send a message that triggers web search (e.g. "wie wird das Wetter in Münster heute")
3. Observe that the AI's internal reasoning, system prompt instructions, and pipeline metadata are visible in the response bubble

---

## Security impact

- **System prompt reverse-engineering** — any user can read the full internal instructions
- **Architecture disclosure** — reveals how search results are injected into the prompt
- **Prompt injection surface** — blurred boundary between user input and system-injected content (search results in `user` role) could be exploited to override instructions

---

## Notes

- This is a **regression** — was fixed before, now reappeared
- Likely causes:
  - Web search results injected as `user`-role message instead of `tool`/`system`
  - Model (likely reasoning model) outputs chain-of-thought as plain text instead of `<thinking>` tags
  - Backend does not strip reasoning output before sending to frontend
- Frontend has handling for `type: 'thinking'` parts (filtered from content) — but reasoning arrives as regular `type: 'text'`
- Could be related to: system prompt change, model switch, or response parsing regression in backend

---

## Screenshots/Logs

Image

cristiangrxs:
## Summary
Structured `reasoning` stream chunks from thinking models (e.g. gpt-oss) were rendered as regular answer text in the multitask pipeline, Again regeneration, media generation and the OpenAI-compatible API — leaking internal instructions and pipeline details to end users (#1067).

## Changes
- Add `App\AI\Stream\StreamChunk::visibleText()` — extracts only user-visible answer text from provider stream chunks and drops `reasoning`/`finish` chunks
- `ChatRunner` (multitask): stop accumulating/streaming reasoning chunks into the task card and the assembled reply (the leak shown in the issue screenshot)
- `AgainHandler`: drop the generic `content` fallback that swallowed reasoning chunks on regeneration
- `MediaGenerationHandler`: non-streaming accumulator only keeps content chunks
- `OpenAICompatibleController`: streaming endpoint no longer forwards reasoning chunks as `delta.content`
- `ChatHandler`: inject web search results into the **system** prompt instead of the user turn (both streaming and non-streaming paths; user-message fallback kept for models without system-message support) and reword the block so the model treats it as system-provided reference data that never overrides instructions — closes the prompt-injection surface called out in the issue

## Verification
- [ ] Not tested (explain)
- [ ] Manual
- [x] Tests added/updated

New regression tests: `StreamChunkTest` (7 cases), `RunnersTest::testChatRunnerDropsReasoningChunks` (replays the exact chunk sequence from the issue screenshot), `ChatHandlerTest::testHandleInjectsWebSearchResultsIntoSystemPrompt`.

Full gate run locally: backend lint ✅, backend tests ✅ (1918 passed), frontend lint ✅, `vue-tsc` ✅, frontend tests ✅ (610 passed). PHPStan reports 52 pre-existing `class.notFound` errors in `TritonProvider.php` (missing local gRPC stubs) that are identical on a clean `main` checkout and unrelated to this change.

## Notes
Fixes #1067. The legacy single-message path (`StreamController`) already separated reasoning correctly via `<think>` wrapping — only the four consumers above leaked. No frontend changes needed; the existing `<think>`-part filtering keeps working.

## Screenshots/Logs
Leak reproduced in the issue (multitask task card + assembled reply):
> "User wrote 'moi'. They want translation into English. … The instruction: translate the user text into 'en'. Return ONLY the translation. So output 'me'.me"

After the fix only the answer text (`me`) reaches the stream — verified by the replayed chunk sequence in `RunnersTest::testChatRunnerDropsReasoningChunks`.

### DIFF
--- a/backend/src/AI/Stream/StreamChunk.php
+++ b/backend/src/AI/Stream/StreamChunk.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\AI\Stream;
+
+/**
+ * Helper for interpreting provider stream chunks.
+ *
+ * Providers emit either plain strings or structured arrays
+ * (['type' => 'content'|'reasoning'|'finish', ...]) — see
+ * {@see \App\AI\Interface\ChatProviderInterface}. Consumers that accumulate
+ * user-visible text MUST go through visibleText() so internal
+ * chain-of-thought ('reasoning') and control signals ('finish') never leak
+ * into the rendered response (issue #1067).
+ */
+final class StreamChunk
+{
+    private function __construct()
+    {
+    }
+
+    /**
+     * Extract the user-visible answer text from a stream chunk.
+     *
+     * Plain string chunks are visible text. Structured chunks contribute only
+     * when their type is 'content'; untyped arrays carrying a content key are
+     * treated as content for backward compatibility with older producers.
+     *
+     * @param string|array<string, mixed> $chunk
+     */
+    public static function visibleText(string|array $chunk): string
+    {
+        if (!is_array($chunk)) {
+            return $chunk;
+        }
+
+        if ('content' !== ($chunk['type'] ?? 'content')) {
+            return '';
+        }
+
+        $content = $chunk['content'] ?? '';
+
+        return is_string($content) ? $content : '';
+    }
+}
--- a/backend/src/Controller/OpenAICompatibleController.php
+++ b/backend/src/Controller/OpenAICompatibleController.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\Controller;
 
 use App\AI\Service\AiFacade;
+use App\AI\Stream\StreamChunk;
 use App\Entity\User;
 use App\Repository\ModelRepository;
 use App\Service\ModelConfigService;
@@ -306,12 +307,11 @@ class OpenAICompatibleController extends AbstractController
                             return;
                         }
 
-                        $content = '';
-                        if (is_array($chunk)) {
-                            $content = $chunk['content'] ?? '';
-                        } elseif (is_string($chunk)) {
-                            $content = $chunk;
-                        }
+                        // Only visible answer text is forwarded — reasoning
+                        // chunks (chain-of-thought) are never exposed (#1067).
+                        $content = is_string($chunk) || is_array($chunk)
+                            ? StreamChunk::visibleText($chunk)
+                            : '';
 
                         if ('' === $content) {
                             return;
--- a/backend/src/Service/Message/AgainHandler.php
+++ b/backend/src/Service/Message/AgainHandler.php
@@ -2,6 +2,7 @@
 
 namespace App\Service\Message;
 
+use App\AI\Stream\StreamChunk;
 use App\Entity\Message;
 use App\Entity\MessageMeta;
 use App\Entity\User;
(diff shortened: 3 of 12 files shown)


FExB17 (approved):

FExB17:
## Code review notes

The fix is clean and well-tested. Two minor gaps worth noting for future hardening (not blockers for this PR):

1. **`WhatsAppService` accumulator consistency** — `WhatsAppService::handleIncomingMessage()` (line ~782) has its own chunk accumulator that manually checks `type === 'content'`. It works correctly today (reasoning is filtered), but it doesn't use `StreamChunk::visibleText()`. If a new chunk type is added in the future, this accumulator won't benefit from the centralized filter. Consider migrating it to `visibleText()` in a follow-up.

2. **`MessageForwardingService` can forward `<think>` blocks** — When a user continues a WhatsApp-originated chat in the web UI, `StreamController` persists `$responseText` including `<think>…</think>` wrapped reasoning. `MessageForwardingService::forwardIfNeeded()` then forwards this raw text to WhatsApp. The `<think>` tags would be visible as literal text in the WhatsApp message. This is an edge case (web-to-WhatsApp forward, not direct WhatsApp reply) but worth a follow-up fix.

Neither affects the core fix for #1067 — the four main leak paths (ChatRunner, AgainHandler, MediaGenerationHandler, OpenAICompatibleController) and the web-search prompt injection surface are all properly addressed.

cristiangrxs:
Both follow-ups are addressed in 68916508b:

1. **WhatsAppService accumulator** — now uses the centralized `StreamChunk::visibleText()` filter instead of its own type check, so it automatically benefits from any future chunk-type handling.
2. **MessageForwardingService** — `forwardIfNeeded()` strips `<think>…</think>` blocks (including unterminated ones) via `AiResponseSanitizer::stripForDisplay()` before forwarding to WhatsApp, and skips the forward entirely when no visible text remains.

Covered by two new regression tests in `MessageForwardingServiceTest`. Full gate re-run: lint ✅, PHPStan ✅ (now 0 errors), backend suite ✅ (1920 tests).

PR #1076, opened 2026-06-12
fix(chat): reconcile streamed message with persisted state after SSE complete (#1070)

### LINKED ISSUE #1070: Streaming: eliminate dual-path divergence for file/media rendering
## Summary

Generated files and media (audio, images, video, documents) are repeatedly invisible during live SSE streaming but appear correctly after page reload. This is not one bug — it is a structural problem caused by two independent, unsynchronized code paths that resolve the same message content.

---

## Problem / Motivation

The frontend renders message content from two completely separate sources depending on timing:

1. **Live (SSE streaming):** `ChatView.vue` accumulates SSE events (`data`, `file`, `audio`, `task_file`, `complete`) and mutates `message.parts` / `message.files` / `taskPlan.cards` in-place during the stream.
2. **After reload:** `history.ts` → `GET /api/v1/chats/{id}/messages` rebuilds `parts` and `files` from the persisted `Message` entity (`m.file`, `m.files[]`).

These paths have different logic, different data shapes, and different code. Every new feature (multitask DAG, WhatsApp, email, widget) must implement both paths correctly and keep them in sync. This is never enforced and consistently breaks.

The pattern has caused at least 15 distinct bugs across 4 months:

- #374 — Again button after image-gen failure
- #501 — Widget takeover messages invisible
- #583, #603, #624 — Model/metadata badges refresh-only
- #625 — MEDIAMAKER audio player missing live
- #626 — Email/widget/WhatsApp generated media: text only on web
- #955 — WhatsApp voice notes as plain text
- #976 — WhatsApp audio 404 / crash
- #1018, #1022 — Human takeover every 2nd message only
- #1055 — Generated documents gone after reload
- #1057 — DAG assembled answer text dropped live

Current instance: TTS audio in a DAG turn is invisible until reload because `taskPlan.active` suppresses the `audio` SSE handler, but the backend does not emit the audio as a `task_file` event.

Each past fix was a point-repair patching one specific gap between the two paths (`pushMediaPart`, `persistTaskPlanFiles`, `extractMediaParts`, ...). None reduced the surface area for this class of bug. `ChatView.vue` is now ~2400 lines of SSE dispatch that every feature branch extends.

---

## Goal

A single authoritative source for the final rendered state of a message, so that streaming and reload can never diverge. The class of "visible only after reload" bugs should be structurally impossible, not just caught by tests.

---

## Acceptance criteria

- [ ] After SSE `complete`, the final message state (parts, files, metadata, taskPlan) is derived from one authoritative source — not from accumulated SSE events alone
- [ ] Adding a new file-producing capability (e.g. a new DAG node type) does not require touching the SSE accumulation logic in `ChatView.vue` to make the file render correctly
- [ ] Audio, image, video, and document output is visible both during live streaming AND after page reload — for single-bubble and multitask DAG turns
- [ ] An E2E test covers: send a multi-step DAG prompt (text + TTS), assert media player visible without reload, reload the page, assert media player still visible
- [ ] The current TTS-audio-in-DAG bug is resolved as part of this work, not as a separate patch

---

## Notes

related: #374, #501, #583, #603, #624, #625, #626, #955, #976, #1018, #1022, #1055, #1057

Key files:
- `frontend/src/views/ChatView.vue` — SSE dispatch (~2400 lines, the main accumulation surface)
- `frontend/src/stores/history.ts` — reload path (`loadMessages`)
- `frontend/src/utils/mediaParts.ts` — #625 race fix (documents the problem well)
- `backend/src/Controller/StreamController.php` — SSE emission + persistence
- `backend/src/Service/Multitask/Execution/DagExecutor.php` — `task_file` emission

Recommended approach: on `complete`, refetch the finalized message from the backend (`GET /messages/{id}`) and reconcile it with the streamed state, treating the persisted version as authoritative for files/media/metadata. Streaming stays responsible for live text chunks, progress indicators, and task-card animations — but the **final** rendered content comes from one place.

cristiangrxs:
## Summary
Generated media (e.g. TTS audio in a multitask/DAG turn) was invisible during live streaming and only appeared after a page reload; the persisted backend row is now refetched and reconciled after SSE `complete`, making it the single authoritative source for the final message state.

## Changes
- Backend: extract message serialization from `ChatController` into a new `MessageApiFormatter` service — single source of truth for the API row shape
- Backend: add `GET /api/v1/messages/{id}` (ownership-checked, full OpenAPI annotations) returning the exact same shape as the chat history endpoint
- Frontend: new `messageMapper.ts` — shared `mapApiMessageRow` used by both the reload path (`history.loadMessages`) and the new post-stream reconciliation, so the two paths can no longer diverge
- Frontend: `history.reconcileMessage` + `reconcileLocalMessage` — after SSE `complete`, missing media parts (image/video/audio) are appended to the live bubble, media already shown by task cards is not duplicated, and persisted metadata (files, aiModels, webSearch, searchResults, topic, …) wins when present
- Frontend: `ChatView.vue` triggers reconciliation in both `complete` handlers (initial stream + continuation)
- TestProvider: `synthesize()` now writes a real silent MP3 frame into the upload dir instead of returning a non-existent `/tmp` path, so voice-reply turns work in dev/test stacks
- Tests: PHPUnit (formatter shape, auth/ownership/404 on the new endpoint), Vitest (mapping + reconciliation rules), and a new E2E test asserting the TTS player in a DAG turn is visible live AND after reload

## Verification
- [ ] Not tested (explain)
- [ ] Manual
- [x] Tests added/updated

Full pre-commit gate is green: backend lint, PHPStan (0 errors), PHPUnit (1913 tests), frontend lint, `vue-tsc`, Vitest (618 tests). E2E: new #1070 test plus `multitask`, `chat`, `chat-again`, `chat-bubble-monotonic`, `task-prompts` specs pass against the CI-like test stack (port 8001).

## Notes
Closes #1070. Approach follows the recommendation in the issue (refetch + reconcile after SSE `complete`). Frontend schemas were regenerated from the updated OpenAPI spec (`make -C frontend generate-schemas` produced no diff outside the generated aliases).

## Screenshots/Logs
E2E acceptance (CI-like test stack):
```
✓ multitask.spec.ts › a multi-node request renders task cards that complete
✓ multitask.spec.ts › TTS audio in a DAG turn is visible live and after reload (#1070)
```

### DIFF
--- a/backend/config/services.yaml
+++ b/backend/config/services.yaml
@@ -705,6 +705,8 @@ services:
 when@dev:
     services:
         App\AI\Provider\TestProvider:
+            arguments:
+                $uploadDir: '%kernel.project_dir%/var/uploads'
             tags:
                 - { name: 'app.ai.chat' }
                 - { name: 'app.ai.embedding' }
@@ -722,6 +724,8 @@ when@test:
             alias: App\Realtime\Publisher\NullPublisher
 
         App\AI\Provider\TestProvider:
+            arguments:
+                $uploadDir: '%kernel.project_dir%/var/uploads'
             tags:
                 - { name: 'app.ai.chat' }
                 - { name: 'app.ai.embedding' }
--- a/backend/src/AI/Provider/TestProvider.php
+++ b/backend/src/AI/Provider/TestProvider.php
@@ -14,6 +14,11 @@ class TestProvider implements ChatProviderInterface, EmbeddingProviderInterface,
 {
     private const FAKE_TOKENS_PER_EMBED = 8;
 
+    public function __construct(
+        private readonly string $uploadDir = '/var/www/backend/var/uploads',
+    ) {
+    }
+
     public function getName(): string
     {
         return 'test';
@@ -483,7 +488,21 @@ class TestProvider implements ChatProviderInterface, EmbeddingProviderInterface,
     // TextToSpeechProviderInterface
     public function synthesize(string $text, array $options = []): string
     {
-        return '/tmp/test_audio.mp3';
+        // Write a real (silent) MP3 frame into the upload dir root —
+        // AiFacade::moveToUserPath() expects the returned filename to
+        // exist there so it can move it into the user path. Returning a
+        // non-existent path made every voice-reply turn on the test
+        // provider fail silently in the move step (issue #1070 E2E).
+        if (!is_dir($this->uploadDir)) {
+            mkdir($this->uploadDir, 0775, true);
+        }
+
+        $filename = 'tts_test_'.uniqid().'.mp3';
+        // Single silent MPEG-1 Layer III frame (128 kbit/s, 44.1 kHz).
+        $frame = "\xFF\xFB\x90\x64".str_repeat("\x00", 413);
+        file_put_contents($this->uploadDir.'/'.$filename, $frame);
+
+        return $filename;
     }
 
     public function synthesizeStream(string $text, array $options = []): \Generator
--- a/backend/src/AI/Provider/TestProvider.php
+++ b/backend/src/AI/Provider/TestProvider.php
@@ -209,6 +209,19 @@ class TestProvider implements ChatProviderInterface, EmbeddingProviderInterface,
             ], JSON_THROW_ON_ERROR);
         }
 
+        // web_search + chat plan — used by @webSearch E2E tests.
+        if (str_contains($text, 'websearch:')) {
+            return json_encode([
+                'version' => 1,
+                'language' => 'en',
+                'reply_node' => 'n2',
+                'tasks' => [
+                    ['id' => 'n1', 'capability' => 'web_search', 'inputs' => ['query' => '$message.text']],
+                    ['id' => 'n2', 'capability' => 'chat', 'depends_on' => ['n1'], 'inputs' => ['text' => '$n1.text']],
+                ],
+            ], JSON_THROW_ON_ERROR);
+        }
+
         return json_encode([
             'version' => 1,
             'language' => 'en',
--- a/backend/src/Controller/ChatController.php
+++ b/backend/src/Controller/ChatController.php
@@ -6,9 +6,9 @@ use App\Entity\Chat;
 use App\Entity\User;
 use App\Repository\ChatRepository;
 use App\Repository\MessageRepository;
-use App\Repository\SearchResultRepository;
 use App\Service\File\DataUrlFixer;
 use App\Service\File\OgImageService;
+use App\Service\Message\MessageApiFormatter;
 use App\Service\WidgetSessionService;
 use Doctrine\ORM\EntityManagerInterface;
 use OpenApi\Attributes as OA;
@@ -27,10 +27,10 @@ class ChatController extends AbstractController
         private EntityManagerInterface $em,
         private ChatRepository $chatRepository,
         private MessageRepository $messageRepository,
-        private SearchResultRepository $searchResultRepository,
         private WidgetSessionService $widgetSessionService,
         private DataUrlFixer $dataUrlFixer,
         private OgImageService $ogImageService,
+        private MessageApiFormatter $messageApiFormatter,
         private LoggerInterface $logger,
     ) {
     }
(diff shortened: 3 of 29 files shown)


FExB17:
## QA finding (manual test against this branch, real Groq planner)

**TL;DR:** the media half of #1070 works, but the dual-path divergence the issue is actually about is *not* closed for **multitask/DAG turns** — before vs. after reload still look different.

### What works (the named acceptance criterion)
TTS audio in a DAG turn is visible **live** and **survives reload**. In the DB the audio is persisted on every turn (`BFILE=1`, `BFILETYPE=audio`), and the `reconcileMessage` → `mapApiMessageRow` path renders it without a reload. 👍

### What still diverges (the issue's broader goal)
For a DAG turn, **only the reply node's text is persisted** as the message body (`ResultAssembler::assemble` → `compose_reply`), plus files. The per-node **task cards** (e.g. the `summarize` node holding the full answer) are **live-streaming-only** — `mapApiMessageRow` has no card data, so a reload cannot rebuild them.

Result:
- **Live:** rich task cards + full plan + audio
- **After reload:** just the `compose_reply` text + audio → the full plan is gone

Concrete DB evidence (same prompt, two DAG turns):

| BID | persisted `BTEXT` | audio |
|-----|-------------------|-------|
| 44  | full training plan | yes |
| 46  | `…zusammenfassen: $n1.text` | yes |

The root cause is structural: the issue recommends "treat the persisted row as authoritative", but for DAG turns the persisted row is **poorer** than the live state (compose_reply only, no cards). `reconcileLocalMessage` correctly leaves live text untouched (otherwise it would wipe the cards), but on reload there is no live state, so only the impoverished persisted version remains. Hence the divergence persists.

To actually satisfy #1070 for DAG turns, the backend needs to either (a) persist + rebuild the full DAG render state (per-node cards), or (b) guarantee `compose_reply` always contains the complete final answer (not a summary/reference).

### Separate, pre-existing bug surfaced during testing (not introduced by this PR)
BID 46 above persisted the literal `$n1.text`. The input resolver only resolves a value when the **whole** string is a reference (`NodeContext::resolve` → `str_starts_with($value, '$')` + strict regex). When the planner emits prose **with an embedded** `$nX.text` (instead of a pure reference), it is passed through verbatim and ends up in the saved reply. This is intermittent (model-dependent on Groq `gpt-oss-120b`) and unrelated to the streaming/reload work here — worth its own issue.

### Note on testing
The deterministic E2E (`multitask.spec.ts › #1070`) only exercises the media path, so it stays green even though the DAG text path still diverges. Manual testing with a real planner is what exposes it.

cristiangrxs:
Both issues you flagged are addressed in the latest commit (`ba2dbe295`).

**DAG card divergence (main goal)**

`ResultAssembler::assemble` now builds a `task_plan_render` metadata key — a `{reply_node, cards[]}` structure with one entry per visible node (`uiKind() !== 'hidden'`). Each card carries `nodeId`, `capability`, `kind`, `state`, `text`, and `url`/`type` from the first file — the same data the live `task_file` SSE event carries. `StreamController` persists this as `task_plan` in `MessageMeta` (JSON, no migration needed) next to the existing `multitask=1`. `MessageApiFormatter` decodes it and includes it in both the history and single-message endpoints. On the frontend, `mapApiMessageRow` reconstructs a `TaskPlanState` (with `active: false`) and `ChatMessage.vue` now renders `TaskPlanBubble` whenever `taskPlan.cards.length > 0`, not only when streaming. After a reload the full plan — all cards with their final `done`/`failed`/`skipped` state and body text — is visible again.

**Embedded `$n1.text` resolver bug**

`NodeContext::resolve` now distinguishes pure references (the entire string is a single `$…` token, detected by a strict `^…$` regex) from prose strings with embedded tokens. For the prose case a `preg_replace_callback` interpolates every recognised `$nX.text`/`$message.text`/`$message.fileText` occurrence in-place. Unknown or unresolvable tokens become an empty string so no placeholder leaks into the persisted reply.

**Test coverage added**

- `ResultAssemblerTest`: card shape, partial-failure card state, skipped-node state
- `NodeContextTest`: embedded single/multiple/message tokens, unresolvable token → empty string, pure-reference path unchanged
- `MessageControllerTest`: `taskPlan` round-trips through the formatter
- `messageMapper.spec.ts`: card rebuild from API row, card-media dedup against flat parts, `reconcileLocalMessage` keeps live `taskPlan` untouched
- `multitask.spec.ts` E2E: after `page.reload()` both `task-card-n1` and `task-card-n2` must still carry `data-state=done`

Full gate (backend lint/phpstan/1935 tests, frontend lint/types/623 tests) passes.

FExB17 (approved):

FExB17:
## Manual QA Review — UX Feedback

Tested the multi-task (DAG) flow with web search + image generation and web search + summary.

### 1. Sources dropdown inconsistency in DAG vs single-task

**Problem:** When a DAG turn includes a `web_search` node, the search results are dumped as raw plain text in both the task card and the message body (with URLs inline as text). In contrast, a single-task web search or certain DAG configurations (web_search + summary) correctly show the **Sources (N)** dropdown carousel at the bottom of the message.

This happens because the `MessageApiFormatter` only populates the structured `searchResults[]` array when `web_search_query` meta is set on the OUT message — and the `ResultAssembler` doesn't consistently propagate that meta for all DAG configurations.

**Expected:** DAG turns with a web_search node should produce the same Sources dropdown UX as single-task web search.

### 2. Task card text for `search` kind — no markdown rendering

The `TaskCard.vue` only uses `MessageText` (markdown) for `isProseKind` (kinds `text`, `extract`). The `search` kind falls through to a plain `<p>{{ card.text }}</p>`, so structured search output with URLs shows without any formatting.

This is borderline acceptable since search results are mostly structured plain text, but if we keep showing them in the card body, at minimum the URLs should be clickable. Alternatively: don't show the raw search text inside the card at all (it's redundant with the message body below).

### 3. Task cards after completion — good UX

The persisted task cards on reload are a clear UX win. The user can trace what happened step-by-step even after refreshing. The only nitpick: the `search` card body shows the entire 10-result raw dump, which is very long. Consider either:
- Collapsing the search card body with a "show details" toggle
- Only showing the result count + query in the card, with full results in the message body

cristiangrxs:
All three UX points addressed in the latest commit (`f145f7d95`).

**1. Sources dropdown in DAG turns**
Root cause: `ResultAssembler` only forwarded the reply node's metadata, so `search_results` from the `web_search` node was silently dropped. Fixed in two layers:
- `ResultAssembler.assemble()` now lifts `search_results` from the first successful `web_search` node into the top-level metadata.
- `StreamController` (both streaming and non-streaming paths) now computes `effectiveSearchResults = result['search_results'] ?? response['metadata']['search_results']`, so DAG turns set `web_search_query`/count on the OUT message and populate the Sources carousel in the `complete` event.
- `WebSearchRunner` injects `SearchResultRepository` and persists fresh results to the IN message (the `reused_prefetched` branch is skipped — `MessageProcessor` already saved those).

**2 & 4. Search card body — compact summary instead of raw dump**
`ResultAssembler` now stores `{query, resultsCount}` on search cards instead of the full formatted text. `DagExecutor` emits the same fields in the `done` `task_update` event for live parity. `TaskCard.vue` renders `"Searched the web · N sources"` (i18n-keyed for all four locales); old persisted cards without the new fields fall back to the kind label.

Unit tests: `ResultAssemblerTest` (+3), `RunnersTest` (+2), `messageMapper.spec.ts` (+1). An `@webSearch` E2E test is included but not tagged `@ci` since it requires a live BraveSearch key.

PR #1095, opened 2026-06-15
fix(chat): support asking about uploaded video content via transcription

### LINKED ISSUE #722: Chat: Cannot ask questions about uploaded file content (e.g. video); previously had search/RAG-style
## Problem

I can upload files (e.g. a video), but I **cannot ask about the file’s contents in chat** and get a meaningful answer. There used to be a workflow where uploaded material could be queried (similar to semantic / RAG search); **that dedicated “Semantic Search” area no longer exists**, and the gap is **not** covered in chat either.

---

## Expected

1. Upload a file (e.g. video) to **Files** and attach or reference it in **Chat**.
2. Ask questions about what’s **in** that file (summary, facts, “what does it say”, etc.).
3. The assistant should answer using **processed content** from the file (extracted text, transcript, or equivalent).

## Actual

- Chat does **not** ground answers in the file (e.g. video yields no usable text path, or “unsupported file type” / generic replies asking for the document).
- The **older semantic-search-style experience** is gone from the UI, and **nothing equivalent** in chat replaces it for “questions about this file’s content.”

---

## Steps to reproduce

1. Upload a file (e.g. MP4) via **Files** (or attach from library in chat).
2. In **Chat**, ask a concrete question about the file’s content (e.g. “summarize this”, “what is the main point”).
3. Observe: no real use of file content, or explicit rejection / “please provide text” behavior.

---

## Notes

- **Product change:** “Semantic Search” as a separate area **no longer exists**; expectation was that **chat + files** would still allow content Q&A.
- Likely related: preprocessing supports docs/audio/images for text; **video** may not be converted to text anywhere on the path to the model.

---

## Screenshots/Logs

Image

cristiangrxs:
## Summary
Uploaded videos (e.g. MP4) can now be queried in chat: their audio track is transcribed and the assistant answers/summarises from that transcript instead of replying "This file type cannot be analyzed" (#722).

## Changes
- `MessagePreProcessor`: add `VIDEO_EXTENSIONS` and route video through the same extraction pipeline as documents (transcript stored in `BFILETEXT`), for both the multi-file and legacy single-file paths.
- `FileAnalysisHandler`: recognise video, answer/summarise from the transcript via the document path, and surface clear pending/failed messages while transcription is in progress or yields no speech.
- `MessageClassifier`: force messages with a video attachment to the `FileAnalysisHandler` route so video-only messages are no longer answered as text-less chats (methods renamed to `*AnalyzableNonImage*`).
- Reuses the existing `FileProcessor` + `WhisperService` (ffmpeg audio extraction + Whisper) pipeline — no new dependencies.

## Verification
- [ ] Not tested (explain)
- [x] Manual
- [x] Tests added/updated

Backend gate (matches CI):
- `make -C backend lint` — pass
- `make -C backend phpstan` — pass (0 errors)
- `make -C backend test` — pass (1923 tests, 0 failures)

New unit tests in `FileAnalysisHandlerMultiFileTest`: video happy path (summary via document path), still-transcribing (pending), and no-speech (failed).

> Note: an end-to-end manual upload of a real MP4 through the UI was not performed in this environment; the change is covered by unit tests and reuses the already-tested video transcription pipeline.

## Notes
Closes #722. Only backend PHP files changed, so frontend checks are not required per the contribution guide.

## Screenshots/Logs
Before (from the issue): a summarisation request on an uploaded MP4 returned "This file type cannot be analyzed. Supported types include documents..., images..., and audio files...".

### DIFF
--- a/backend/src/Service/File/FileProcessor.php
+++ b/backend/src/Service/File/FileProcessor.php
@@ -48,6 +48,15 @@ final readonly class FileProcessor
     ];
 
     /**
+     * Video container formats that must always have their audio track stripped
+     * before being sent to an external STT API — even though some containers
+     * (mp4, mkv) appear in API_SUPPORTED_AUDIO_FORMATS. A full video file can
+     * easily be 40+ MB while the extracted audio-only payload is a few MB;
+     * every hosted Whisper endpoint enforces a ~25 MB ceiling.
+     */
+    private const VIDEO_EXTENSIONS = ['mp4', 'mov', 'avi', 'mkv', 'mpeg', 'mpg'];
+
+    /**
      * Audio formats supported by external APIs (OpenAI/Groq Whisper).
      * Formats not in this list need to be converted before sending.
      *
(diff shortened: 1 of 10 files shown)


FExB17:
## QA Review — Manual end-to-end test with real MP4 upload

Tested the video pipeline with a real 43 MB WhatsApp MP4 (2:27 min, English speech). The routing, handler logic, classifier changes, and tests all look solid. However, the underlying `FileProcessor` transcription pipeline has two gaps that make video transcription fail silently for most real-world uploads.

---

### 1. Full video file is sent to external STT APIs instead of the extracted audio track

Both configured external providers reject the file:

```
Groq:   Audio file too large: 41.2MB (max 25MB)
OpenAI: 413: Maximum content size limit (26214400) exceeded (26376956 bytes read)
```

The ~25 MB upload ceiling is an industry-wide constraint across hosted Whisper endpoints and will apply to any future STT provider. The root cause is in `FileProcessor::ensureApiCompatibleFormat()`: since `mp4` is listed in `API_SUPPORTED_AUDIO_FORMATS`, the **full 43 MB video** (including the video track) is sent as-is. The actual audio payload is only ~4.5 MB when extracted via `ffmpeg -vn`.

Videos should always have their audio track stripped and compressed (mono 16 kHz MP3) before being sent to any external endpoint — regardless of whether the container format is technically "supported". This brings the file well under any provider's size limit and also reduces latency and cost. The fix belongs in `FileProcessor` (not in individual provider implementations) so every current and future provider benefits automatically.

### 2. No fallback to local Whisper when external providers fail

`FileProcessor::extractFromAudio()` treats external and local Whisper as mutually exclusive paths. When a `SOUND2TEXT` model is configured, `hasConfiguredSttProvider()` returns `true` and local Whisper.cpp is never attempted — even when the external call fails:

```php
if ($this->aiFacade->hasConfiguredSttProvider($userId)) {
    return $this->extractFromAudioExternal($absolutePath, $baseMeta, $userId);
}
```

I verified that **local Whisper works perfectly** on the same file — full accurate transcript in ~35 seconds. A robust strategy would be: try external provider → on failure, fall back to local Whisper.cpp.

### 3. `FileUploadService` sets status `extracted` with 0 text instead of `error`

When `FileProcessor::extractText()` returns empty text (because the API rejected the file), `FileUploadService::processFile()` sets the status to `extracted`, not `error`. This means the `video_transcription_failed` route added by this PR is never reached for standalone file uploads — the file appears "extracted" with no content, and the user gets no error indication.

Note: `MessagePreProcessor` (changed in this PR) correctly sets `status: 'error'` when extraction returns empty text, so the **chat attachment path is fine** — it's only the **standalone file upload path** through `FileUploadService` that has this gap.

### 4. Minor: Missing `files.status_error` i18n key in `en.json` and `de.json`

Pre-existing, not introduced by this PR: `files.status_error` exists in `tr.json` and `es.json` but is missing from `en.json` and `de.json`. When a file hits the `error` status, the UI shows the raw key instead of a translated string.

---

### Summary

| # | Finding | Severity | Scope |
|---|---------|----------|-------|
| 1 | Video sent to API without stripping video track → exceeds ~25 MB limit | Critical | `FileProcessor` — provider-agnostic |
| 2 | No fallback to local Whisper when external STT fails | Critical | `FileProcessor` — provider-agnostic |
| 3 | `FileUploadService` silent success on failed extraction | Medium | Standalone upload path |
| 4 | Missing i18n keys for `status_error` | Low | Pre-existing |

Items 1 and 2 are the blockers: without them, any video upload >25 MB (which is most real-world videos) will silently fail when an external STT provider is configured, even though local Whisper handles it fine.

related: #1097

FExB17 (approved):

PR #1116, opened 2026-06-18
feat(chat): add quote selection ("Mention in chat")

cristiangrxs:
## Summary
Adds a Claude-style "Mention in chat" quote feature: select text in any chat message, click a floating button, and a shortened quote chip is attached above the composer to use as a reference.

## Changes
- **Composable `useMessageQuoting.ts`** (new): detects text selections inside `[data-quotable]` elements, positions a floating button over the selection, exposes `pendingQuote` / `confirmQuote` / `clearPendingQuote`, cleans up listeners on unmount. Plus `formatQuoteAsBlockquote` helper.
- **`QuoteSelectionButton.vue` / `QuoteChip.vue`** (new): teleported floating button centered above the selection (flips below near the viewport top) and a truncated, removable quote chip.
- **Main chat**: `ChatView` + `ChatInput` wire the button/chip and pass `quotedText` / `quotedMessageId` through the send flow (auth + guest). The quote is rendered as a blockquote in the user bubble.
- **Backend (AI main chat only)**: `StreamController` reads + validates the params (same chat/user), stores them in `processingOptions`, and persists them via `MessageMeta` (survives reload, no migration). `ChatHandler::formatQuotedReferenceForPrompt()` injects a dedicated "Quoted reference" context block into the system prompt (streaming + non-streaming + no-system-role fallback). `ChatController` serializes the quote from `getMeta()` for history.
- **Operator surfaces**: `LiveSupportView` and `WidgetSessionsView` reuse the same composable/button/chip; the quote is prepended to the outgoing message as a Markdown blockquote (human-to-human, no AI binding).
- **i18n**: new `chat.quote.*` keys in en/de/es/tr.

## Verification
- [ ] Not tested (explain)
- [ ] Manual
- [x] Tests added/updated

Full gate green locally:
- `make -C backend lint` · `make -C backend phpstan` (No errors)
- `make -C backend test` — 1974 passed
- `make -C frontend lint` · `npm run check:types` (vue-tsc)
- `make -C frontend test` — 624 passed
- New tests: `useMessageQuoting`, `QuoteChip`, `QuoteSelectionButton` (Vitest); `ChatHandler::formatQuotedReferenceForPrompt` (PHPUnit). Routing characterization snapshots unchanged.
- `make -C frontend generate-schemas` produced no schema drift.

## Notes
- Persistence uses the existing `MessageMeta` key-value store — no Doctrine migration needed.
- Shared chat is read-only (selection only, no quote→send).
- One active quote at a time (replaceable).

## Screenshots/Logs

### DIFF
--- a/backend/src/Controller/StreamController.php
+++ b/backend/src/Controller/StreamController.php
@@ -37,8 +37,16 @@ use Symfony\Component\Security\Http\Attribute\CurrentUser;
 #[OA\Tag(name: 'Messages')]
 class StreamController extends AbstractController
 {
-    /** Server-side cap for a quoted-reference excerpt ("Mention in chat"). */
-    private const MAX_QUOTED_TEXT_LENGTH = 4000;
+    /**
+     * Server-side cap for a quoted-reference excerpt ("Mention in chat").
+     *
+     * The quote travels as a query parameter on the SSE (GET) stream URL, so an
+     * over-long excerpt could exceed reverse-proxy request limits (e.g. Nginx's
+     * 8 KB default). Kept in sync with the frontend cap (MAX_QUOTE_LENGTH in
+     * useMessageQuoting.ts); 1000 chars is generous for a reference point and
+     * stays URL-safe even worst-case encoded.
+     */
+    private const MAX_QUOTED_TEXT_LENGTH = 1000;
 
     public function __construct(
         private EntityManagerInterface $em,
--- a/backend/src/Controller/StreamController.php
+++ b/backend/src/Controller/StreamController.php
@@ -37,6 +37,9 @@ use Symfony\Component\Security\Http\Attribute\CurrentUser;
 #[OA\Tag(name: 'Messages')]
 class StreamController extends AbstractController
 {
+    /** Server-side cap for a quoted-reference excerpt ("Mention in chat"). */
+    private const MAX_QUOTED_TEXT_LENGTH = 4000;
+
     public function __construct(
         private EntityManagerInterface $em,
         private AiFacade $aiFacade,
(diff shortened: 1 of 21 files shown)


backend/src/Service/Message/Handler/ChatHandler.php
```diff
@@ -323,6 +326,14 @@ public function handle(
             ]);
         }
 
+        // Quoted reference ("Mention in chat") — non-streaming/channel parity.
+        if (!empty($options['quoted_text'])) {
+            $systemPrompt .= $this->formatQuotedReferenceForPrompt($options);
+            $this->logger->info('ChatHandler: Quoted reference appended to system prompt (non-streaming)', [
+                'quoted_message_id' => $options['quoted_message_id'] ?? null,
+            ]);
+        }
```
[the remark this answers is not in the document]
cristiangrxs: Fixed in a6e4ca8a6: the quote is now injected after the system-message capability check (guarded by `null !== $systemPrompt`) and falls back to the user turn via `buildMessages()` (appended to BTEXT) for models without system-message support, mirroring the web-search flow.

backend/src/Controller/StreamController.php
```diff
@@ -295,6 +309,11 @@ public function streamMessage(
         // and we normalize the empty string to null so "no folder selected"
         // behaves the same as an absent parameter.
         $ragGroupKey = $request->query->getString('ragGroupKey') ?: null;
+        // Quoted reference the user selected from an earlier message ("Mention
+        // in chat"). `getString()` rejects array input with a clean 400, and we
+        // normalize the empty string to null so an absent quote behaves the same.
+        $quotedText = $request->query->getString('quotedText') ?: null;
+        $quotedMessageId = $request->query->get('quotedMessageId');
```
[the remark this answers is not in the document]
cristiangrxs: Fixed in a6e4ca8a6: `quotedMessageId` is now read via `getInt()` (rejects array input instead of coercing to 0/1), and `quotedText` is trimmed and capped server-side (MAX_QUOTED_TEXT_LENGTH = 4000) before persisting to meta and adding to processing options.

backend/src/Controller/StreamController.php
```diff
@@ -295,6 +312,19 @@ public function streamMessage(
         // and we normalize the empty string to null so "no folder selected"
         // behaves the same as an absent parameter.
         $ragGroupKey = $request->query->getString('ragGroupKey') ?: null;
+        // Quoted reference the user selected from an earlier message ("Mention
```
FExB17, line 323: **URL length risk**: `quotedText` is sent as a query parameter. At `MAX_QUOTE_LENGTH=4000` the URL-encoded string can reach ~12 KB, which exceeds default URL limits on some reverse proxies (Nginx default: 8 KB, Cloudflare: 16 KB). Consider moving to a POST-based SSE endpoint or storing the quote server-side before streaming.

FExB17 (commented):
Code review: two non-blocking suggestions for future improvement.

backend/src/Controller/StreamController.php
```diff
@@ -295,6 +312,19 @@ public function streamMessage(
...
+        // in chat"). `getString()` rejects array input with a clean 400. We trim
```
FExB17, line 324: **Weak integer validation**: `$request->query->get("quotedMessageId")` returns a string. The downstream `(int) $quotedMessageId` cast silently turns `"abc"` into `0`, which could match a real message id. Safer: `filter_var($quotedMessageId, FILTER_VALIDATE_INT)` or `ctype_digit()` guard before casting.

FExB17 (approved):

backend/src/Controller/StreamController.php
cristiangrxs: Good catch. The SSE stream uses an EventSource (GET-only), so a POST-based endpoint would be a larger architectural change, out of scope here. As a proportionate fix I lowered the cap to 1000 chars on both client and server (kept in sync, with a visible ellipsis when a selection is shortened). Worst-case URL-encoded that stays ~3 KB, well under Nginx's 8 KB default — while remaining generous for a reference point. A POST-based/stored-quote approach remains a sensible future enhancement if we ever need longer excerpts. Fixed in 822f5c5e5.

backend/src/Controller/StreamController.php
cristiangrxs: This was already hardened earlier in the PR: quotedMessageId now uses `$request->query->getInt()`, which validates via `FILTER_VALIDATE_INT` + `FILTER_NULL_ON_FAILURE` and throws a 400 (BadRequestException) on non-integer or array input ("abc", "5abc", `quotedMessageId[]=…`) instead of silently coercing to 0. That's stronger than a plain cast, and 0/absent collapses to null.

Issue #1146, opened 2026-06-19
Billing: cancelled requests not recorded in BUSELOG — cost budget bypass


FExB17:
## Bug

When a user cancels a streaming request (Stop button, page reload, or navigation), the `recordUsage()` call in `StreamController` is **never reached**. This means:

1. The provider-side cost is incurred (video rendered, image generated, tokens consumed)
2. The cost is NOT recorded in `BUSELOG`
3. The user's cost budget (`checkCostBudget()`) is not decremented
4. The user can immediately start another request

**This is an exploitable billing bypass.**

## Severity: P0 (Security/Financial)

A malicious user could:
1. Start a Higgsfield video generation ($2.50/video)
2. Immediately click Stop
3. Video continues rendering on Higgsfield (see #1145)
4. Cost is NOT written to BUSELOG → budget counter stays at $0.00
5. Repeat indefinitely → platform pays thousands, user pays nothing

## Affected Capabilities

| Capability | Provider Cost | Bypass Window |
|-----------|-------------|---------------|
| **VIDEO** (Higgsfield) | ~$2.50/video | 100% — render always completes (5 min) |
| **VIDEO** (Google Veo) | Variable (credits) | 100% — render always completes (1-3 min) |
| **VIDEO** (fal.ai) | Variable (credits) | 100% — render always completes |
| **IMAGE** (OpenAI, Google) | ~$0.04/image | High — generation completes in 5-30s |
| **CHAT** (all LLM providers) | $0.001-0.05/request | Low — streaming stops quickly on disconnect |
| **TTS** (OpenAI, Piper) | $0.001-0.01/request | Medium — synchronous, completes in <5s |

## Root Cause

In `StreamController.php`, the `recordUsage()` calls (lines 1409, 1434) are placed **after** `processStream()` returns successfully. When a client disconnect triggers `RuntimeException('Client disconnected')`, the catch block at line 1707 does:

```php
} catch (\Exception $e) {
    if ('Client disconnected' === $e->getMessage()) {
        $this->logger->info('Stream stopped - client disconnected', [...]);
        return; // ← SILENTLY EXITS — no cost recording!
    }
}
```

The `saveCancelled` endpoint (line 2648) also does **not** call `recordUsage()` — it only persists the message with status "cancelled".

### Option A: Record cost before starting generation (pre-charge)
- Calculate expected cost BEFORE calling the provider
- Write a provisional BUSELOG entry
- If generation fails/is cancelled, mark it as refunded

[...]

Issue #1145, opened 2026-06-19
Video generation: Stop/Cancel does not abort remote renders — credits wasted (all polling providers)


FExB17:
## Bug

When a user clicks Stop (global or per-card) during video generation, the backend fails to cancel the remote render job. The video continues generating on the provider's infrastructure, consuming credits/money, while the user sees nothing.

**Severity: P1 — real money is burned on every cancelled video.**

## Affected Providers

| Provider | Cancel API available? | Currently implemented? | Estimated cost per lost render |
|----------|----------------------|----------------------|-------------------------------|
| **Higgsfield** (DoP, Kling) | ✅ `POST /cancel` | ⚠️ Partially — `cancel_check` exists but is bypassed by `connection_aborted()` | ~$2.50/video |
| **Google Veo** | ✅ `POST {operationName}:cancel` | ❌ Not implemented | Google AI credits (variable) |
| **HuggingFace/fal.ai** (LTX, Hunyuan) | ✅ `PUT /requests/{id}/cancel` | ❌ Not implemented | fal.ai credits (variable) |

### Problem 1: `connection_aborted()` kills process before `cancelRemote()` fires (Higgsfield)

In `StreamController.php`, `ignore_user_abort(false)` is set (line 451). When the client disconnects:
- The streaming callback detects disconnect and throws `RuntimeException('Client disconnected')`
- This terminates the entire processStream execution immediately
- The Higgsfield `pollUntilTerminal()` never gets a chance to call `cancelRemote()`
- The `cancel_check` closure (which would trigger `cancelRemote`) never fires because the process is dead

### Problem 2: Google Veo ignores `cancel_check` entirely

`GoogleProvider::pollVideoUntilComplete()` accepts only `$progressCallback` — it does not read or use `cancel_check` from `$options`. The polling loop has **no abort condition** for cancellation. Even when `MediaCancellationStore` is flagged, Veo keeps polling for up to 600 seconds.

### Problem 3: HuggingFace/fal.ai ignores `cancel_check` entirely

`HuggingFaceProvider::pollUntilComplete()` has **no cancellation check** at all. It polls fal.ai queue status until completion or timeout with no way to interrupt it.

### Problem 4: `sendSSE()` swallows disconnect silently

[...]

Issue #1144, opened 2026-06-19
Multitask: DAG video node cannot use upstream image node output as reference


FExB17:
## Summary

When the planner creates a multi-step plan where a media node depends on the file output of a preceding node (e.g. "generate an image, then animate it" or "generate an image, then edit it"), the downstream node cannot use the upstream file output as its reference. This affects all media-to-media chains, not just video.

---

## Problem / Motivation

- Users naturally expect "create a picture of X, then animate it" or "create a logo, then make it blue" to work as a chain.
- The planner CAN express this dependency (`n2.inputs.image = "$n1.file"`, `dependsOn: ["n1"]`), and `NodeContext::resolveInputs()` already resolves `$nX.file` to the upstream node's first file descriptor.
- BUT `MediaGenerationRunner` only reads `inputs['prompt']` / `inputs['text']` — it never checks for resolved file inputs and never attaches them to the synthetic message.
- `MediaGenerationHandler` detects PIC2PIC and IMG2VID exclusively from files attached to the message (`collectAttachedImagePaths($message)`). Since the synthetic message only carries `$context->message->getFiles()` (user uploads), upstream node outputs are invisible.

**Affected scenarios:**
1. **Image → Video** (IMG2VID): "Generate an image, then animate it as video" → video node has no reference image → falls back to TEXT2VID (Veo) instead of IMG2VID (Higgsfield DoP/Kling)
2. **Image → Image-Edit** (PIC2PIC): "Generate a logo, then make it blue" → edit node has no reference image → falls back to plain TEXT2PIC instead of PIC2PIC editing

---

## Goal

A downstream media node that depends on an upstream node's file output should receive that file as a reference, triggering the correct path (IMG2VID or PIC2PIC) automatically.

---

## Acceptance criteria

- [ ] `MediaGenerationRunner` resolves file-typed inputs (e.g. `inputs['image']` = `$n1.file`) and attaches them to the synthetic message so `MediaGenerationHandler::collectAttachedImagePaths()` finds them
- [ ] Image→Video chain: when a plan has `n2: video_generation, dependsOn: ["n1"], inputs: { image: "$n1.file" }`, the video node uses n1's generated image as IMG2VID reference → Higgsfield DoP/Kling is used
- [ ] Image→Image-Edit chain: when a plan has `n2: image_generation, dependsOn: ["n1"], inputs: { image: "$n1.file" }`, the edit node uses n1's image as PIC2PIC reference
[...]

Issue #1141, opened 2026-06-19
Multitask: per-card Stop button does not unblock input when it is the only running node


FExB17:
## Problem

The per-card Stop button on a running media task card sets the card to "cancelled" locally but does not reliably send the cancel request to the backend, leaving the SSE stream open and blocking new messages.

---

## Expected

Clicking the per-card Stop button on the only running media card should:
1. Set the card to "cancelled" state (works)
2. Send `POST /api/v1/messages/cancel-node` to the backend (sometimes fails)
3. The provider poll aborts, the stream sends "complete", and the input unblocks

---

## Actual

1. The card visually shows "Stopped" (correct)
2. The console logs `⚠️ No currentTrackId – skipping backend notification` — the `cancelTask` API call is never sent because `currentTrackId` is `undefined`
3. The SSE stream stays open, `isStreaming` remains true, and the user sees "Please wait for the current response to finish"
4. Only the global Stop button (chat input) eventually closes the stream (by aborting the EventSource), but even that logs the same "No currentTrackId" warning

---

## Steps to reproduce

1. Start a single-node multitask video generation (e.g. "Erstelle ein kurzes Video: Ein Boot auf einem See")
2. Wait for the video card to show "Working..." with the Stop button
3. Click the per-card Stop button on the video card
4. Observe: card shows "Stopped", but the input remains disabled ("Please wait...")
5. Open browser console: `⚠️ No currentTrackId – skipping backend notification`

---

## Notes

- Root cause: `currentTrackId` in `ChatView.vue` is `undefined` at the time `handleTaskCancel` runs (line 3032). This variable is set at the start of streaming (line 1780) but may be cleared by a race condition or may never have been set if the stream started on a different code path.
- The global Stop works because it calls `streamingAbortController.abort()` + `stopStreamingFn()` which physically close the EventSource — this is independent of `currentTrackId`.
- The `MediaCancellationStore` (backend) is correctly wired: if the cancel request DOES reach the backend, the Higgsfield poll loop checks it every 3s and aborts. The issue is purely frontend: the request never leaves the browser.
- Merged in PR #1130, now on `main` at commit 183c37a43.
- related: feat(multitask): live media progress, per-card Stop, and Higgsfield duration limits

---

## Screenshots/Logs
Mache zwei Sachen: Erstelle ein Bild von einem Sonnenuntergang am Meer. Erstelle ein Video von fliegenden Vögeln über Bergen.
Console output when clicking per-card Stop:
```
[...]

PR #1147, opened 2026-06-19
Fix/higgsfield cancel

orgaralf:
## Summary
4 big issue fixes

## Changes
Caused and triggered by the Higgsfield integration.

## Verification
- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated

### DIFF
--- a/backend/migrations/Version20260429220000.php
+++ b/backend/migrations/Version20260429220000.php
@@ -89,7 +89,18 @@ final class Version20260429220000 extends AbstractMigration
     {
         // Skip on installs that never ran the widget feature (fresh provisioning before
         // 20260417 baseline, or a custom deployment that disabled widgets).
-        if (!$schema->hasTable('BWIDGET_SESSIONS') || !$schema->hasTable('BMESSAGES')) {
+        //
+        // Guard via information_schema rather than the injected Schema API: reading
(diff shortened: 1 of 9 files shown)


orgaralf:
## Fixes

- **Fixes #1146** — Billing bypass on cancel/disconnect. Media usage is now recorded in `MediaGenerationHandler` for both successful and `ProviderCancelledException` paths (`record_media_usage` option + `usage_recorded` flag prevents double-counting), and partial chat cost is recorded best-effort in `StreamController::saveCancelled`.
- **Fixes #1145** — Stop/Cancel now aborts remote video renders. A `cancel_check` callable is threaded through the `GoogleProvider` (Veo) and `HuggingFaceProvider` (fal.ai) polling loops, which call the provider cancel endpoints (`operations:cancel` / `requests/{id}/cancel`) and throw `ProviderCancelledException` instead of wasting credits.
- **Fixes #1144** — DAG video node can now use an upstream image node's output. `MediaGenerationRunner` collects resolved `$nX.file` paths into `reference_image_paths`, `MediaGenerationHandler` normalises and attaches them (IMG2VID/PIC2PIC), and the `tools:plan` prompt documents the chaining pattern.
- **Fixes #1141** — Per-card Stop button now unblocks input. The streaming `trackId` is persisted on `taskPlan` so the right node is cancelled, and the streaming turn is finished locally when the stopped card is the last running one.

PR #1250, opened 2026-07-03
fix(chat): stop dropping system prompts for models without system-role support (office maker hotfix)

orgaralf:
## Summary
- **Bug:** Document generation (officemaker) returned an empty/prose answer instead of a DOCX whenever the chat model was **GPT-5.5 Pro** (IDs 206/207). Root cause: the model JSON has `supportsStreaming: false` and the legacy "no streaming ⇒ no system messages" heuristic in `ChatHandler::modelSupportsSystemMessages()` misread that as "no system role" — the **entire system prompt (topic prompt, RAG context, language directive) was then silently dropped**, so the model never saw officemaker's `{"BFILEPATH":...,"BFILETEXT":...}` contract.
- **Catalog fix:** `ModelCatalog` entries 206/207 now carry `supportsSystemMessages: true` — the Responses API takes system prompts via `instructions`; only streaming is unsupported. Reseed (`app:seed`) propagates it to `BMODELS` via the catalog fingerprint.
- **Defense in depth:** when a model genuinely rejects the system role (e.g. o1 via chat completions), `ChatHandler` now folds the system prompt into the **first user turn** (both `handle()` and `handleStream()`) instead of discarding it — which is what the code comment always claimed but never did.

## Test plan
- [x] 3 new regression tests in `ChatHandlerTest`: fold-into-user-turn (non-streaming + streaming) and explicit-flag override keeps the system role.
- [x] `make -C backend lint` / `make -C backend phpstan` / `make -C backend test` (2467 tests) — all green.
- [x] Live API verification on local stack: "Erstelle mir ein Word-Dokument mit einem kurzen Gedicht über Katzen." with GPT-5.5 Pro as chat model → `__FILE_GENERATED__:katzengedicht.docx`, valid OOXML zip (7.5 KB) downloadable via `/api/v1/files/uploads/...` (before the fix: prose answer, no file).
- [ ] After deploy: reseed models (`app:seed`) so live `BMODELS` rows 206/207 pick up the flag, then re-test document generation on live.

Frontend untouched (backend-only change).

### DIFF
--- a/backend/src/Model/ModelCatalog.php
+++ b/backend/src/Model/ModelCatalog.php
@@ -861,6 +861,11 @@ class ModelCatalog
                 'params' => ['model' => 'gpt-5.5-pro'],
                 'features' => ['reasoning', 'vision'],
                 'supportsStreaming' => false,
+                // Responses API takes system prompts via `instructions` — only
+                // streaming is unsupported. Without this flag the legacy
+                // "no streaming ⇒ no system messages" heuristic silently drops
+                // every topic prompt (broke officemaker document generation).
+                'supportsSystemMessages' => true,
                 'meta' => [
                     'api' => 'responses',
                     'context_window' => '1050000',
@@ -890,6 +895,7 @@ class ModelCatalog
                 'params' => ['model' => 'gpt-5.5-pro'],
                 'features' => ['reasoning', 'vision'],
                 'supportsStreaming' => false,
+                'supportsSystemMessages' => true,
                 'meta' => [
                     'api' => 'responses',
                     'supports_images' => true,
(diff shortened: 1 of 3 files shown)


orgaralf:
Re the three Copilot comments on the mock callbacks: not applicable. PHP closures accept surplus arguments without error — `ArgumentCountError` fires only for *missing* required parameters. Verified empirically: the three tests run green (22 assertions), and a closure declared `function(array $a)` invoked with 3 args executes fine. This is also the established pattern in this file (see `testHandleInjectsWebSearchResultsIntoSystemPrompt`, which mocks `AiFacade::chat()` with the same 1-parameter callback and has been green in CI for months). Capturing only `$messages` is intentional — it's the only argument these tests assert on.

PR #1282, opened 2026-07-07
feat(backend): rolling, condensing conversation summary for long chats

orgaralf:
## Summary
Adds a rolling, tiered, condensing conversation summary that is injected into the chat system prompt so long threads keep their topic / the user's position while the combined context (verbatim recent turns + summary) stays inside a configurable 10–15k character memory window.

## Changes
- **New `ConversationSummaryService`** (`backend/src/Service/Message/`): builds the rolling context from the full chat history.
  - Keeps the newest turns **verbatim** (up to a configurable char budget); condenses everything older into a summary.
  - **Gradient/tiered compression**: the older span is split into recency tiers — the oldest tier is condensed the hardest, the tier nearest the verbatim window the least (exactly the "older condensed more, recent condensed less" behaviour requested).
  - Caches the summary per stable older-span so consecutive turns don't re-summarize; never throws into the turn (any failure falls back to the standard windowed history).
- **`ConversationSummaryConfigService` + `ConversationSummaryConstants`**: BCONFIG-backed settings (group `CONVERSATION_SUMMARY`, owner 0) with code-level defaults, so it works with zero seeded rows. The combined window is clamped to the **10 000–15 000** char band. Master `ENABLED` toggle (default on).
- **`RollingSummaryResult`**: value object carrying the summary + the recent verbatim turns.
- **`ModelConfigService::getSummaryModelConfig()`**: resolves the condensing model `DEFAULTMODEL.SUMMARY → SORT → CHAT`. Configurable (an operator can point it at e.g. a GPT-OSS-120B model); **defaults to the sorting model** as requested.
- **Wiring**: `MessageProcessor.processStream()` builds the rolling context before inference and passes it via `options['conversation_summary']` (trimming the replayed thread to the recent window); `ChatHandler` folds it into the SYSTEM prompt next to memories/feedback (also carried into the first user turn for models without system-message support).

See `_devextras/planning/20260707-rolling-conversation-summary.md` for the full design.

## Verification
- [x] Tests added/updated
- [x] Not tested (explain)

Unit tests were written for the service (disabled / null-chat / short-history no-ops, gradient condensing with verbatim recent kept, model options passed through, provider-failure fallback, per-span caching), the config clamping, the model resolution chain, and the ChatHandler system-prompt injection. **The test/lint/PHPStan gate could not be executed in this environment** — it has no Docker, PHP, or Composer runtime, so `make lint && make -C backend phpstan && make test` was not run here. Please run the standard pre-commit gate before merge.

## Notes
- Follows the exploration finding that the live chat path had no rolling summarization — this fills that gap. Retrieval-based memory (Qdrant memories/feedback/RAG) is untouched and complementary.
- Wired into the **streaming** chat path (the main flow). The non-streaming `process()`/email path can be wired later with the same option; `ChatHandler` already reads `conversation_summary` in the shared prompt-building code.
- BCONFIG defaults are bootstrap-only; because the config service falls back to constants, no migration is needed to ship defaults. To change a default on existing installs later, ship an UPDATE migration.

## Screenshots/Logs

<div><a href="https://cursor.com/agents/bc-a81d542f-722b-4df3-a48c-4e8815f93556"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a81d542f-722b-4df3-a48c-4e8815f93556"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


FExB17:
## Review notes

**1. Full history DB query on every streaming turn (performance)**

`MessageProcessor.processStream()` calls `findAllByChatId()` on every turn — loading the *entire* chat history from the DB — even when the cache hit means no AI call happens. For long chats (100+ messages, which is exactly what this feature targets), that's a non-trivial query per turn.

Consider checking whether summarization is even needed *before* loading the full history — e.g. by comparing the current message count or last message ID against the cached span, or by gating on a cheap count query first.

**2. Null-safety gap in `renderMessage`**

```php
$fileText = $msg->getFileText();
if ('' !== $fileText) {
```

If `getFileText()` returns `null` (which entity getters can do for nullable columns), the condition `'' !== null` is `true` and you get a broken `[attached ...]` fragment. Two lines above, `getText()` is already cast with `(string)` — apply the same here:

```php
$fileText = (string) $msg->getFileText();
```

PR #1333, opened 2026-07-14
ci: move off-critical-path jobs from paid to free runners

FExB17:
## Summary

Follow-up to #1321. Queue-time analysis of the last 12 CI runs showed the paid `ubuntu-latest-m` pool is the only place where jobs wait: free runners always start within ~3s (p90 4s, max 6s across 90 jobs), while `-m` jobs hit up to **81s provisioning latency** (p90 68s) when several of them are requested at once. In the worst observed case that delayed `frontend-build` — the head of the critical path — by 25s because `backend`/`frontend-checks` had grabbed the warm runners first.

The 16 cores only pay off where a parallel workload sits on the critical path:

| Job | Parallel workload? | On critical path? | Runner |
| --- | --- | --- | --- |
| `frontend-build` | yes (two concurrent Vite builds) | **yes** | stays on `ubuntu-latest-m` |
| `backend` | partially (PHPStan yes, PHPUnit is single-process, setup is core-independent) | no (done ~min 2-3, e2e runs to ~min 7) | → `ubuntu-latest` |
| `frontend-checks` | barely (vue-tsc single-threaded, Vitest small) | no | → `ubuntu-latest` |

Effects:
- `frontend-build` now reliably gets the warm `-m` runner at t=0 (no more provisioning-latency outliers on the critical path).
- `backend`/`frontend-checks` run ~30-60s longer on 4 cores — irrelevant, they keep minutes of slack before the e2e jobs finish.
- Paid runner minutes drop by roughly two thirds (~5.5 → ~2 16-core-minutes per run).

Comments in the workflow document the reasoning and when to revisit (e.g. if `backend` ever becomes the longest job).

## Test plan

- [x] `actionlint` — no new findings (two fewer: moved jobs now use a known label)
- [x] CI run on this PR: all green in 7m02s, `frontend-build` queue 3s (run 29344924346)
- [x] `backend` on the free runner: 2m33s (was ~2m on 16 cores) — still finishes ~3.5 min before the e2e jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,9 +57,13 @@ jobs:
 
   backend:
     name: Backend (PHP/Symfony)
-    # 16-core org larger runner (billed per minute) — PHPStan + PHPUnit are the
-    # longest serial stage, so this is where the extra cores pay off.
-    runs-on: ubuntu-latest-m
+    # Free runner on purpose: this job is off the critical path (done around
+    # minute 2-3 while e2e runs until ~minute 7), PHPUnit is single-process
+    # anyway, and free runners start instantly (queue p90 ~4s vs up to ~80s
+    # provisioning on the ubuntu-latest-m pool). Keeping it off the -m pool
+    # also leaves the warm -m runner to frontend-build, which IS the
+    # critical path. Revisit if this job ever becomes the longest one.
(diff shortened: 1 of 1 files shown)


orgaralf (approved):
Good, the right tool for the right job!

PR #1380, opened 2026-07-27
fix: restore document and multitask chat workflows

### LINKED ISSUE #1228: Chat: Bilder können nicht in generierte DOCX-Dokumente eingefügt werden (officemaker)
## Problem

Wenn ein Nutzer im Chat sagt „Füge mein Profilbild in die Bewerbung ein" (oder ein angehängtes Bild in ein zuvor generiertes DOCX eingefügt werden soll), passiert das **nicht**. Der Chat-Dokumentgenerator (`officemaker`) kann nur **Text-Inhalte** (Markdown) in DOCX-Dateien schreiben — Bilder werden komplett ignoriert.

## Reproduktion

1. Nutzer generiert eine Bewerbung als DOCX im Chat (→ funktioniert)
2. Nutzer hängt ein Bild an und sagt „Füge dieses Bild als Profilbild in die Bewerbung ein"
3. **Erwartet:** Das DOCX enthält das Bild
4. **Tatsächlich:** Das Bild wird ignoriert, die AI antwortet nur mit Text oder generiert ein DOCX ohne Bild

### 1. `officemaker`-Prompt kennt keine Bilder

Der Prompt in `PromptCatalog::officeMakerPrompt()` (Zeile 807–898) instruiert die AI, ein JSON mit `BFILEPATH` + `BFILETEXT` zurückzugeben. `BFILETEXT` ist **reiner Markdown-Text**. Es gibt keine Möglichkeit, auf angehängte Bilder zu verweisen oder diese einzubetten:

```
{"BFILEPATH":"bewerbung.docx","BFILETEXT":"# Bewerbung\n\nSome body text."}
```

Der Prompt erwähnt weder Bilder noch ein Bildformat.

### 2. `DocumentGeneratorService::writeDocx()` kann keine Bilder einbetten

Die Methode (Zeile 78–125) konvertiert Markdown → HTML → PhpWord via `WordHtml::addHtml()`. Obwohl `normalizeVoidTags()` ``-Tags normalisiert (für XML-Validität), werden **keine Bilddateien aufgelöst oder eingebettet**. Ein `` im HTML hat keinen Zugriff auf das Dateisystem.

### 3. `ChatHandler::storeGeneratedFile()` arbeitet nur mit Text

Die Methode (Zeile 1918–1997) schreibt den AI-Content über `documentGenerator->write()`. Es wird nur `$content` (String/Markdown) übergeben — kein Mechanismus für Bildanhänge.

### 4. Bilder werden zwar gesehen aber nicht weitergereicht

Der `ChatHandler` kann Bilder über `extractImageDataUrls()` an Vision-Modelle senden — das Modell **sieht** das angehängte Bild. Aber der officemaker-Prompt weiß nicht, was damit zu tun ist, und `DocumentGeneratorService` hat keine API um Bilddaten in ein DOCX einzubetten.

### 5. Datei-Bearbeitungskontext ist nur Text

Beim Bearbeiten eines zuvor generierten DOCX wird der gespeicherte `fileText` (Markdown-Quelltext) in den Prompt injiziert (Zeile 1208–1217). Es gibt kein Konzept von „das DOCX enthält auch folgende Bilder".

## Erwartetes Verhalten

1. Nutzer kann im Chat ein Bild anhängen und die AI bitten, es in ein generiertes DOCX einzufügen
2. Das Bild wird tatsächlich als eingebettetes Bild im Word-Dokument platziert
3. Auch bei Folge-Bearbeitungen bleibt das Bild erhalten

### Phase 1: Bild-Platzhalter im officemaker-Prompt

1. **Prompt erweitern:** `officemaker`-Prompt ergänzen, damit die AI auf angehängte Bilder verweisen kann, z.B. über einen Marker wie `{{IMAGE:file_id}}` oder `` im Markdown
2. **BFILETEXT-Format erweitern:** Neben dem Text-Content auch Bild-Referenzen unterstützen (z.B. als zusätzliches JSON-Feld `"images"` oder Inline-Marker)

### Phase 2: DocumentGeneratorService mit Bild-Support

3. **`writeDocx()` erweitern:** Bild-Platzhalter im HTML/Markdown erkennen und über PhpWord's `$section->addImage()` als echte eingebettete Bilder einfügen
4. **Bildquellen auflösen:** Angehängte Dateien über ihre `File`-Entity und den Dateipfad laden, Base64-Data-URLs oder Dateipfade in OOXML-Bilder konvertieren

### Phase 3: Bearbeitungs-Kontext

5. **Bild-Metadaten im `fileText` persistieren:** Beim Speichern des generierten DOCX auch Bild-Referenzen merken, damit Folge-Bearbeitungen die Bilder beibehalten

## Betroffene Dateien

- `backend/src/Prompt/PromptCatalog.php` — `officeMakerPrompt()` um Bild-Support erweitern
- `backend/src/Service/File/DocumentGeneratorService.php` — `writeDocx()` um Bild-Einbettung erweitern
- `backend/src/Service/Message/Handler/ChatHandler.php` — Bild-Anhänge an den officemaker-Pfad durchreichen
- `backend/src/Service/Multitask/Execution/Runner/DocumentGenerationRunner.php` — DAG-Pfad ebenfalls anpassen

### LINKED ISSUE #1229: Multitask: Interactive Task Cards — collapse prose, keep media visible, add copy/download/lightbox
## Summary

Redesign task cards after DAG completion: prose cards with redundant text collapse to eliminate duplication (#1129), cards with unique content stay visible, media cards get proper interaction controls (copy, download, lightbox). Replaces the fragile `dedupeTaskPlanProse` workaround with a clean UX pattern.

Supersedes / closes: #1129

---

## Problem

1. **Text duplication (#1129):** After DAG completion, prose tasks (chat, summarize, translate) often show their full text inside the card AND again below as compose_reply — redundant wall of text.
2. **Overly aggressive collapse would hide unique content:** Sometimes compose_reply is just a short connector sentence ("Hier ist das Gedicht und ein Dokument...") while the actual content lives only in the card. Collapsing everything would hide it.
3. **No interaction controls:** Task cards lack copy buttons, download buttons, and image lightbox — unlike regular chat message parts which have all of these.
4. **Images not clickable:** Generated images inside task cards are plain `` tags with no way to view them larger.
5. **No download affordance for media:** Audio/video/image cards have no download button (only document cards have one).

---

### Smart collapse: only collapse when card text is redundant

There are two distinct scenarios:

| Scenario | compose_reply content | Card text | Correct behavior |
|---|---|---|---|
| **A — Duplication** (#1129) | Same as card text (pass-through) | Redundant | **Collapse card** — compose_reply is primary |
| **B — Unique content** (screenshot) | Short connector/summary | Unique content | **Keep card visible** — it has the actual content |

#### Collapse decision logic

**Option 1 — Frontend text comparison (simpler, no backend change):**
```
collapsed = card.state === 'done'
  && isProseKind
  && composeReplyContains(card.text)
```
Reuses the existing `dedupeTaskPlanProse` similarity logic but as a collapse trigger instead of a hide hack.

**Option 2 — Backend signals redundancy (cleaner, small backend change):**
ResultAssembler knows whether compose_reply passes through upstream text (`"text": "$n1.text"`) or uses a custom synthesis prompt. It can emit a `redundant: true` flag per card in `task_plan_render`.
```
collapsed = card.state === 'done' && card.redundant === true
```

Recommended: **Option 2** — it's deterministic, no fuzzy text matching, and the backend already has this information.

---

### Visibility rules per card type (after completion)

| Card kind | Collapse rule | Visible state |
|---|---|---|
| **Text** (chat, summarize, translate, extract) | Collapse only if `redundant` | If collapsed: header + Copy btn + expand chevron. If not redundant: full text visible + Copy btn |
| **Image** | Never collapse | Thumbnail + Download + click → Lightbox |
| **Audio** | Never collapse | Mini-player + Download |
| **Video** | Never collapse | Player/thumbnail + Download |
| **Document** | Never collapse | Filename + icon + Download |
| **Search** | Never collapse | Query + source count (already compact) |

**During streaming:** All cards stay expanded with live progress (current behavior, unchanged).

---

### Action buttons (all card types when `done`)

| Action | When shown | Behavior |
|---|---|---|
| **Copy** | Text cards, search cards | Copy card text to clipboard |
| **Download** | Image, audio, video, document cards | Download the generated file |
| **Lightbox** | Image cards (click on thumbnail) | Full-screen image view (reuse MessageImage pattern) |

---

### Cleanup (after Option 2 is implemented)

- Remove `dedupeTaskPlanProse()` from `taskPlanDisplay.ts` (replaced by backend `redundant` flag)
- Remove reload dedup logic from `messageMapper.ts` (#1165 workaround becomes unnecessary)

---

### Backend change (Option 2)

In `ResultAssembler.php`: when building `task_plan_render.cards[]`, compare each text card's content with the final reply node output. If the card text is contained in (or identical to) the compose_reply output, set `"redundant": true` on that card.

### Frontend changes

- `TaskCard.vue`: Add collapsed/expanded state, action buttons (copy, download), image lightbox on click
- `TaskPlanBubble.vue`: Pass compose_reply reference for collapse decision (if Option 1)
- `taskPlanDisplay.ts`: Remove `dedupeTaskPlanProse` (after backend flag ships)
- `messageMapper.ts`: Remove reload dedup logic
- Reuse `MessageImage.vue` lightbox pattern for task card images

### Existing code hooks

- `TaskCard.vue` already has `isProseKind` / `isMediaKind` computed properties
- `MessageImage.vue` has working lightbox (fixed overlay, download, Escape-to-close)
- `ChatMessage.vue` has `copyMessageText` pattern for clipboard
- Card state machine already has `done` state for triggering transitions

---

## Visual reference

Current state (problem visible):

Current UI showing duplication and missing controls

- "Antwort" card shows full poem text (unique content — should NOT collapse here)
- "Dokument" card shows download link (correct, keep visible)
- compose_reply below: "Hier ist das Gedicht..." (short connector — not a duplication)
- Image at top: no lightbox, no download button

---

## Steps to reproduce

**Duplication case (should collapse):**
1. Send "erstelle mir einen Trainingsplan und einen Plan für meine Altersvorsorge"
2. Both text cards show full content AND compose_reply repeats the same text below

**Unique content case (should NOT collapse):**
1. Send "schreibe ein Gedicht über dieses Bild und erstelle ein Dokument daraus"
2. "Antwort" card has the poem, compose_reply only says "Hier ist das Gedicht und das Dokument"
3. Collapsing the card would hide the poem

**Missing controls:**
1. Any DAG with image generation — image not clickable, no download
2. Any completed text card — no copy button

---

## Acceptance criteria

- [ ] Prose cards collapse only when their text is redundant with compose_reply
- [ ] Prose cards with unique content stay visible after completion
- [ ] Media cards (image, audio, video, document) never collapse
- [ ] Copy button on all completed prose/search cards
- [ ] Download button on all completed media/document cards
- [ ] Image click opens lightbox (full-screen view)
- [ ] Streaming behavior unchanged (cards expand with live progress)
- [ ] Backend emits `redundant` flag per card in task_plan_render (Option 2)
- [ ] `dedupeTaskPlanProse` removed after backend flag ships
- [ ] Reload dedup in messageMapper removed
- [ ] No text duplication visible when compose_reply passes through card text
- [ ] Unique card content preserved when compose_reply is just a connector

---

## Related

- Closes #1129 (answer content displayed twice)
- Related: #1081 (card layout changes after reload)
- Related: #1057 (original empty bubble issue that introduced fullContent accumulation)
- Related: #1165 (file_analysis text shown twice after reload)

### LINKED ISSUE #1251: TTS/Audio: generated file stores audio duration (Tika metadata) instead of source text
## Problem
TTS-generated audio files have two compounding issues:

1. **Routing failure (MessageClassifier):** When a TTS-generated file is re-attached to a message, `MessageClassifier::attachedFileIsAnalyzableNonImage()` does not recognize the generic kind `audio` (stored by `GeneratedFileRegistrar`) as an analyzable attachment — because `AUDIO_EXTENSIONS` only lists concrete extensions (`mp3`, `wav`, `ogg`, …). The message falls through to the AI sorter and gets handled as a regular chat turn, so the AI responds as if no file was attached ("Es scheint, als hättest du eine Datei mitschicken wollen – bei mir ist aber nichts angekommen").

2. **Wrong extracted text (FileProcessor):** Even when the file does reach `FileProcessor::extractText()` (e.g. via "Describe, vectorize & sort"), it shows Tika-extracted audio duration (e.g. "41.856") as text content instead of the poem/text that was actually synthesized — because `isTranscribableMedia('audio')` returns false and the file falls through to Tika.

---

## Expected
- A TTS-generated file re-attached to a message is recognized as audio and routed to `FileAnalysisHandler`.
- A TTS-generated file's `BFILETEXT` contains the source text (the content that was synthesized to speech), so the file is searchable by its actual content.

## Actual
1. `MessageClassifier::attachedFileIsAnalyzableNonImage()` reads `$file->getFileType()` → `'audio'`, prefers it over the filename extension, checks `in_array('audio', AUDIO_EXTENSIONS)` → `false`. Message bypasses the `analyzefile` route entirely.
2. `FileProcessor::extractText()` receives `$ext = 'audio'`, `isTranscribableMedia('audio')` → `false`, falls through to Tika which extracts MP3 metadata (duration in seconds) as "text content".

**Note:** Issue #1236 fixed this same generic-kind resolution in `FileAnalysisHandler::resolveFileType()` (falls back to filename extension), but the same fix was never applied to `MessageClassifier::attachedFileIsAnalyzableNonImage()`.

---

## Steps to reproduce
1. In chat, send a message that triggers TTS generation, e.g. "lies mir ein gedicht vor" (read me a poem).
2. Wait for the DAG to complete (text generation → text2sound).
3. In a new message, attach the generated `tts_*.mp3` file and ask "was wird hier gesagt?" (what is said here?).
4. **Result:** AI responds as if no file was attached (routing failure).
5. Alternatively: Open Files, find the generated `tts_*.mp3`, click "Describe, vectorize & sort".
6. **Result:** File text shows a number like "41.856" instead of the source text (wrong extraction).

**Contrast:** Downloading the same `tts_*.mp3` to the PC and re-uploading it works correctly — the upload process assigns `fileType = 'mp3'` (concrete extension) instead of the generic kind `'audio'`.

---

## Notes

**Root cause chain (for the implementer):**

### Problem 1: Routing gap
- `MessageClassifier::attachedFileIsAnalyzableNonImage()` (line ~468): `$fromType = strtolower($file->getFileType())` → `'audio'`. Since it's non-empty, it is used as `$ext`. Then `in_array('audio', AUDIO_EXTENSIONS)` → `false`. Returns `false` → message is NOT classified as `analyzefile`.
- The fix from #1236 (`FileAnalysisHandler::resolveFileType()`) that maps generic kinds to concrete extensions via filename was never replicated here.

### Problem 2: Wrong extracted text
- `Text2SoundRunner::run()` has the synthesized `$text` (the poem) but only returns it as a file descriptor with `path`/`type`/`local_path` — the source text is discarded.
- `GeneratedFileRegistrar::register()` creates the File entity with NO `fileText` and stores `fileType = 'audio'` (generic kind).
- When later processed, `FileProcessor::extractText(…, 'audio')` → `isTranscribableMedia('audio')` → `false`.
- Falls through to Tika, which extracts MP3 metadata (duration in seconds) as "text content".

**Possible fixes (layered):**
1. **Routing fix:** Apply the same generic-kind resolution from `FileAnalysisHandler::resolveFileType()` to `MessageClassifier::attachedFileIsAnalyzableNonImage()` — if `$fromType` is a generic kind (`audio`/`video`/`image`/`document`), fall back to the filename extension before checking the extension lists.
2. **Primary content fix:** Store the source text on `File.fileText` at TTS generation time — the `Text2SoundRunner` already has it; pass it through to the registrar.
3. **Secondary content fix:** `isTranscribableMedia()` should also recognize generic kinds (`'audio'`, `'video'`) or resolve the real extension from `$file->getFileName()`/path before checking.
4. **Tertiary content fix:** Set `messageId` on generated files so `indexPrompt` can trace back to the originating message's assembled text.

---

## Screenshots/Logs
- Attaching a platform-generated TTS file: AI responds "Es scheint, als hättest du eine Datei mitschicken wollen – bei mir ist aber nichts angekommen" (no file recognized).
- Uploading the same file from PC: works correctly (routed to file analysis, transcribed).
- File detail modal showing `tts_*.mp3` with content "41.856" (duration) instead of the source text.

### LINKED ISSUE #1343: Chat: multi-task card content missing after reload until second reload
## Problem

Reloading a chat while a multi-task turn is running shows the task cards (since #1142) but with no text content — only empty shells with status badges. The full content only appears after a second reload (once the turn has completed and the OUT message exists).

---

## Expected

After reloading mid-turn, completed cards should show their response text, running cards should show their current state, and media cards should show any produced URLs — the same content visible during the live SSE session.

## Actual

All cards render with `text: ''` because `BMESSAGE_TASKS` stores only `nodeId`, `capability`, `kind`, and `state` — no text, URL, or progress data. `mapInProgressTurn()` maps every card with `text: ''`. During live streaming, content arrives via SSE `task_chunk` events, but after F5 there is no SSE reconnect or polling. The user is forced to reload a second time after the turn completes.

---

## Steps to reproduce

1. Send a multi-task prompt (e.g. "Do these 3 tasks: 1) Explain Docker 2) List 5 TS advantages 3) Write a haiku")
2. While the task cards are still running, press F5
3. Observe: cards reappear but are empty shells — no text content
4. Wait for the turn to complete, reload again → now content appears

---

## Notes

related: #1142, #1334

Key files:
- Backend: `TaskPlanStore::loadCards()` — only reads `BNODEID, BCAPABILITY, BSTATUS`
- Backend: `InProgressTurnResolver` — passes the card-shell shape without text
- Frontend: `mapInProgressTurn()` in `messageMapper.ts:567` — sets `text: ''` for every card
- Frontend: `history.ts:440` — appends the skeleton bubble, no follow-up polling or SSE reconnect

Fix direction: persist partial card content (text, URL, error) to `BMESSAGE_TASKS` as each node settles, then include it in the `inProgressTurn` API response.

---

## Screenshots/Logs

Cards after first reload (empty shells, status only):
```
Aufgabenplan · 2/4
  Antwort ● Fertig      ← no text
  Antwort ● Fertig      ← no text
  Antwort ♪ Läuft…
  Antwort ♪ Läuft…
```

### LINKED ISSUE #1379: Usage taximeter: planning cost billed but missing from the per-message badge on single-node turns
## Problem

On turns that invoke the multi-task planner but execute through the single-node/legacy path, the PLANNING LLM call is charged in `BUSELOG` yet omitted from the per-message usage badge and `ai_usage_extra`, so the badge total and the session "MODELS USED" list stop reconciling with the ring total.

---

## Expected

The per-message badge, the streamed `usage_extra`, the persisted `ai_usage_extra` meta, and the session model list include the PLANNING step on every turn where a planning call was billed — matching the DAG path and the `BUSELOG` total (the reconciliation intent of #1373 / #1378).

## Actual

`PLANNING` is written to `BUSELOG` with real provider tokens (`BESTIMATED = 0`) but never reaches the usage payload: `complete.usage_extra`, the persisted `ai_usage_extra`, and the reloaded history all carry only the `SORT` and chat `LLM` entries. The badge under-reports the turn by the whole planning cost (~7–8k tokens for a default plan prompt).

Observed on a trivial single-intent `moin` turn (AI-sorted, single-node chat plan):

```
BUSELOG:               SORT 3468 tok · PLANNING 7666 tok · MESSAGES 818 tok
complete.usage_extra:  [ SORT 3468 ]            (PLANNING absent)
badge total:           818 + 3468 = 4286        (~7.7k short)
```

Multi-intent DAG turns are unaffected: there `runDag()` writes `planning_usage` into the assembled metadata directly.

---

## Root cause

`TaskPlanExecutor::withPlanningUsage()` merges the planning entry into `$handlerResult['response']['metadata']`. The single-node/legacy path returns the handler result unchanged through `runSingleNode()` → `InferenceRouter::routeStream()` → `ChatHandler::handleStream()`, whose shape is `['metadata' => [...]]` with no `response` wrapper. `$handlerResult['response']` is therefore `null`, the early guard `if (!is_array($response)) return $handlerResult;` fires, and `planning_usage` is dropped.

Downstream consumers read `planning_usage` from the top-level `metadata` (which is exactly where `runDag()` already places it), so only the DAG path ends up populated.

`TaskPlanExecutorTest::testPlanningUsageSurvivesDelegationToLegacyRouter` mocks `routeStream()` returning `['response' => ['content' => ..., 'metadata' => []]]` and asserts `$result['response']['metadata']['planning_usage']`. That shape is not produced by any real handler, so the test stays green against a structure the code never emits.

All affected call sites share `withPlanningUsage()`: the `shouldUseLegacyRouter` delegate and the `all_failed` DAG fallback, in both `executeStream()` and `execute()` — covering the web, WhatsApp, and email paths.

## Fix direction

Attach the planning entry to the same top-level `metadata` key that the handlers and `runDag()` use, dropping the non-existent `response` indirection:

- In `withPlanningUsage()`, read and write `$handlerResult['metadata']` (initialising it when absent) instead of `$handlerResult['response']['metadata']`.
- Rewrite `testPlanningUsageSurvivesDelegationToLegacyRouter` to return the real handler shape (`['metadata' => [...]]`) and assert `$result['metadata']['planning_usage']`; anchor the mock to the exact `ChatHandler::handleStream()` return shape so it cannot drift again.
- Verify the top-level `metadata` shape survives `MessageProcessor` (`['response' => $response]`) and is picked up by `StreamController::appendContextUsage()`, `WhatsAppService::storeOutgoingMessage()`, and `WebhookController::email()`.

The planning usage must travel inside the usage payload exactly like the DAG path — no display-side re-derivation of the cost.

## Steps to reproduce

1. Enable the usage taximeter (`USAGE_TAXIMETER.ENABLED = 1`); keep `CLASSIFIER.FAST_PATH_ENABLED` at its default (off) so every message is AI-sorted and the planner runs.
2. Send a trivial single-intent message (e.g. `moin`).
3. Compare the streamed `complete.usage_extra` and the persisted `ai_usage_extra` meta on the OUT message against `BUSELOG` for the same turn:

```sql
SELECT BACTION, BPROMPT_TOKENS, BCOMPLETION_TOKENS, BTOKENS, BCOST, BESTIMATED
FROM BUSELOG
WHERE BUSERID = :uid
ORDER BY BID DESC
LIMIT 6;
```

A `PLANNING` row is present in `BUSELOG` but missing from `usage_extra`.

## Verification

- `complete.usage_extra`, the persisted `ai_usage_extra`, and the reloaded `GET /api/v1/chats/{id}/messages` response each include a `PLANNING` entry for single-node planner turns.
- Badge total (chat + sort + planning + any media/transcription) equals the sum of the matching `BUSELOG` rows for the turn.
- Full backend gate green (`make -C backend lint`, `make -C backend phpstan`, `make -C backend test`), including the corrected `TaskPlanExecutorTest`.

---

## Notes

related: #1373, #1378
Key refs: `TaskPlanExecutor::withPlanningUsage()`, `ChatHandler::handleStream()` return shape, `TaskPlanExecutorTest::testPlanningUsageSurvivesDelegationToLegacyRouter`, `StreamController::appendContextUsage()`.

### LINKED ISSUE #1382: Documents: cannot embed a previously generated chat image in a generated document
## Problem
A document generated by the officemaker flow cannot embed an image that was generated earlier in the same conversation; only images attached to the current message or already embedded in the edited document resolve.

---

## Expected
When a user references an image produced earlier in the chat (for example a generated profile photo) and asks to place it in a generated document, the image is embedded. This matches natural user expectation and adjacent flows that already reuse prior-turn content.

## Actual
The AI reports it has no access to the earlier image, or the document node emits an image marker that cannot be resolved and no image is embedded. In the same multitask turn an image node can produce a file while the parallel document node cannot reference it.

Image marker resolution is limited to two sources, neither of which covers a previously generated chat image:
- `{{IMAGE:attached:N}}` — resolved only from files attached to the current message (`DocumentImageReferenceResolver::resolve()`, `backend/src/Service/File/DocumentImageReferenceResolver.php:30`, via `imageAttachments()` reading `$message->getFiles()`).
- `{{IMAGE:file:N}}` — resolved only for images already embedded in the document being edited (`resolvePersistent()`, `backend/src/Service/File/DocumentImageReferenceResolver.php:66`).

A previously generated chat image is neither attached to the new message nor known to the model as a `file:ID`, so it can never be referenced.

Observed in one turn: the image node produced a file while the document node failed to resolve its marker.
```
MediaGenerationRunner: node produced file {"node_id":"n1","path":"/api/v1/files/uploads/01/000/00001/2026/07/377_google_1785156842.jpg","type":"image"}
DocumentGeneratorService: DOCX HTML parsing failed, using plain text fallback {"error":"Document image reference could not be resolved: attached:1"}
```

---

## Steps to reproduce
1. Generate an image in a chat conversation.
2. In a later message ask the officemaker flow to create a DOCX and include that earlier image.
3. The earlier image is not embedded; the model reports no access to it, or the marker fails to resolve.

---

## Notes
related: #1228
Fix direction: expose images generated earlier in the current thread to the model as selectable `file:ID` references and extend the resolver to resolve those thread-scoped generated files. Within a single multitask turn, make the image node output available as an input to a downstream document node.

---

## Screenshots/Logs
See log excerpt above.

cristiangrxs:
## Summary
Fixes document image embedding (including reuse of earlier chat images), restores reliable multitask task-card state, keeps billed planning usage visible on single-node turns, and stops a single Stop click from creating duplicate cancellation rows.

## Changes
- Embed user-owned chat image attachments in generated DOCX files through persistent, validated image markers.
- Expose images from earlier turns and upstream multitask image nodes to the officemaker prompt so generated documents can embed them via `{{IMAGE:file:ID}}` instead of inventing unresolvable `attached:N` markers.
- Strip unresolvable image markers instead of aborting document generation; leave non-DOCX formats without raw markers.
- Preserve embedded images during document edits and missing-binary regeneration, including streaming and multitask generation paths.
- Restore smart multitask task-card data and refresh task plans correctly after chat history reloads.
- Persist one cancellation message per user Stop (do not treat `StreamCancelledException` as a provider ERROR row).
- Preserve and cover generated TTS source text metadata.
- Attach billed planning usage to the real top-level handler metadata on legacy/single-node turns so badges, persisted history, and model totals reconcile.
- Add and update backend and frontend regression tests for these workflows.

## Verification
- [ ] Not tested (explain)
- [ ] Manual
- [x] Tests added/updated
- [x] `make lint`
- [x] `make test`

## Notes
- Closes #1228
- Closes #1229
- Closes #1251
- Closes #1343
- Closes #1379
- Closes #1382
- Do not merge until all required GitHub checks pass. The final merge will be performed manually by the maintainer.

## Screenshots/Logs
Not applicable; verification is covered by the automated backend and frontend regression suites.


FExB17:
## Review findings from local testing

Two defects surfaced while exercising the officemaker + media flow on this branch. Finding 1 is a regression introduced by this PR and blocks the DOCX image feature; finding 2 is pre-existing and observed during the same session.

### 1. Unresolvable image marker aborts the whole DOCX (regression, introduced by #1228 in this PR)

When the model emits `{{IMAGE:attached:1}}` but no image is attached to the current message, document generation fails completely and produces no file.

Repro: officemaker request "write a job application as DOCX and add the profile image", with no image attached to that message.

Logs:
```
ChatHandler: Detected file generation {"filename":"bewerbung_mustermann.docx","extension":"docx","content_length":2143}
DocumentGeneratorService: DOCX HTML parsing failed, using plain text fallback {"error":"Document image reference could not be resolved: attached:1"}
ChatHandler: Failed to write file {"error":"Document image reference could not be resolved: attached:1"}
ChatHandler: File generation failed
```

Root cause: `DocumentImageReferenceResolver::resolve()` leaves the `{{IMAGE:attached:1}}` marker in the content when the current message has no matching attachment (`attached_images: 0`), and adds no path to `$images`. Both write paths in `DocumentGeneratorService` then throw on the unresolved marker:

- `addDocxContent()` — `backend/src/Service/File/DocumentGeneratorService.php:185` throws `Document image reference could not be resolved: attached:1`.
- The catch falls back to `buildPlainTextDocx()` — `backend/src/Service/File/DocumentGeneratorService.php:217` — which contains the same `null === $path` throw, so the fallback re-throws instead of rescuing the document.

The result contradicts the fallback's own intent (comment at `DocumentGeneratorService.php:120`: "so we never ship a blank-but-valid DOCX"). A single unresolvable marker discards the entire document.

Fix direction: strip/skip unresolvable `{{IMAGE:...}}` markers rather than throwing. The plain-text fallback must never fail on an image marker, so a document is always produced (without the image) instead of nothing.

### 2. Single user cancel persists two cancellation rows on multitask/media turns (pre-existing)

Cancelling one multitask/media turn writes two outgoing messages sharing the same tracking id, so the chat shows two "cancelled" cards for one action.

Evidence (`BMESSAGES`, one turn, `BTRACKID` 1785156776223):
```
BID  BDIRECT  BTOPIC  BTEXT
375  OUT      CHAT    ⚠️ Vom Benutzer abgebrochen
376  OUT      ERROR   ## ⚠️ Stream cancelled by user
```
Only one cancel event is logged: `Cancelled message saved {"message_id":375}`.

Two independent paths each persist a cancellation:
- `backend/src/Controller/StreamController.php:3443` — the `/save-cancelled` endpoint writes the CHAT row (375).
- `backend/src/Controller/StreamController.php:1290` — the streaming result-error branch writes the ERROR row (376) because the media node returns `success:false, error:"Stream cancelled by user"`.

On a plain chat turn the cancel is thrown as `StreamCancelledException` and swallowed silently at `StreamController.php:2232`, so only one row is written; the double row is specific to multitask/media turns. These cancel paths are not touched by this PR — candidate for a separate issue.

### Out of scope (separate issue)

Reusing a previously generated chat image in a new document is not supported by #1228 (only current-message attachments and already-embedded `{{IMAGE:file:N}}` markers resolve). Filed separately.

cristiangrxs:
@FExB17 Thanks — both findings are addressed on the branch, and `main` is merged in (the TaskCard conflict is resolved).

**1. Unresolvable image marker aborted the whole DOCX (regression) — fixed**

An image that cannot be embedded now costs only that image, never the document:

- `DocumentImageReferenceResolver::resolve()` drops markers it cannot resolve instead of leaving them in the content. The cleaned text is what gets persisted as `BFILETEXT`, so a dead reference is not carried into the next edit either.
- `DocumentGeneratorService::addDocxContent()` and `buildPlainTextDocx()` skip an unknown reference and log a warning instead of throwing — the plain-text fallback can no longer fail on the very marker that sent it there.
- Same principle applied to a WebP that fails conversion: the image is skipped, the document is written.
- Formats that cannot embed images (`txt`, `md`, `csv`, `xlsx`, `pptx`) no longer leak the raw `{{IMAGE:…}}` marker into the download.

Your repro ("job application as DOCX with the profile image", nothing attached) now produces the document without the image.

**2. Two cancellation rows per Stop click — fixed as well**

You were right about the two paths. A user cancel is now carried as a cancel end to end instead of degrading into an error:

- `AiFacade::chatStream()` and `CircuitBreaker` pass `StreamCancelledException` through untouched. As a side effect, Stop clicks no longer count as provider failures — a few of them could otherwise open the circuit for that provider.
- `MessageProcessor::processStream()` reports the cancel as `cancelled: true` rather than a generic error (returned, not re-thrown, so the WhatsApp/widget/again callers keep their current behaviour).
- `StreamController` ends a cancelled turn silently. Only the `/save-cancelled` row remains, and the incoming message keeps its `cancelled` status instead of being overwritten with `error`.

**Out of scope, unchanged:** reusing a previously generated chat image in a new document stays in the separate issue.

Coverage: updated/added unit tests for the resolver, the DOCX writer (including the plain-text fallback path), `MessageProcessor`, `CircuitBreaker`, `AiFacade` and the controller guard. Full gate green locally (`make lint`, `phpstan`, `make test`, `vue-tsc`).

cristiangrxs:
Follow-up: #1382 (reusing a previously generated chat image in a new document) is now fixed in this PR instead of staying a separate issue.

**Root cause:** resolution was never the blocker — `resolvePersistent()` accepts any user-owned `{{IMAGE:file:ID}}`. The model simply never learned that those images exist, so it guessed `{{IMAGE:attached:1}}` (exactly the log in the issue) and the marker could not resolve.

**Fix:**
- New `DocumentImageCatalog` collects the images a document turn may embed: attachments of the current message, files produced by an upstream multitask node, and the images of the conversation (M2M attachments plus the BFILES rows generated media links through `BMESSAGEID`). Every entry is validated against the uploads dir and ownership.
- The officemaker system prompt now carries that list as the exact markers the resolver accepts back — one source of truth for offered vs. accepted references. An empty catalog prints an explicit "no images available, do not write a marker" instruction, which removes the invented-marker failure mode at the source.
- `DocumentGenerationRunner` forwards the image files of the nodes it depends on, so an image node and a document node in the same multitask turn work together. The planner prompt documents the wiring (`depends_on` + `inputs.images: ["$n1.file"]`) with a worked example; the planner snapshot was re-recorded and reviewed line by line.
- `DocumentImageReferenceResolver` shares the catalog's attachment and path handling instead of duplicating it.

Tests: new `DocumentImageCatalogTest`, `ChatHandlerDocumentImageContextTest`, resolver test for the previously-generated-image case, two runner tests for upstream image forwarding. Full backend gate green (lint, PHPStan, 2774 tests). Frontend untouched.

orgaralf (approved):
lgtm

PR #1411, opened 2026-07-31
fix(ci): cut releases on request and start the app sync with least privilege

cristiangrxs:
## Summary

Two fixes to the release chain.

### Releases are cut on request

Tagging every app-relevant merge was the wrong default:

- The impact classifier is **fail-closed**, so a large share of merges classifies as `store-required`. A tag per merge would produce a continuous stream of review candidates.
- Every merge would burn a patch version, so the tag series stops carrying meaning.
- A `v*` tag starts a **second full CI run on the same commit**, because `ci.yml` builds the version-tagged images on tags.

`release-tag.yml` (renamed from `auto-tag.yml`) now runs on request and collects everything merged since the previous release tag. The classifier needs no change: it already diffs against the last tag. Two guarantees the `workflow_run` trigger gave implicitly are now explicit, because a manual start can name any commit — the ref must be the default branch, and the commit must have a successful `CI` run. A `dry_run` input reports the classification and resulting tag without creating it.

### The synchronization is started, not dispatched

The first live release failed with `403 Resource not accessible by integration`. Creating a repository dispatch event requires **Contents write** on the target, while the dispatcher app was documented as read-only.

Raising it to Contents write is the wrong fix: this key is a secret of a **public** repository, so leaking it would grant push access to the private app repository. The workflow now calls `sync-synaplan.yml` directly, which is gated on **Actions write** — enough to start a release, not enough to change a line of code.

Requires [metadist/synaplan-apps#4](https://github.com/metadist/synaplan-apps/pull/4) to be merged first, and the dispatcher app's permission changed to Actions: read and write.

## Test plan

- [x] `node --test "tests/*.test.mjs"`
- [ ] `gh workflow run release-tag.yml -f dry_run=true` reports a classification and a tag without creating one
- [ ] A real release reaches the sync PR in the app repository

## Added: classifier allow-lists

The fail-closed fallback caught 96 of the 215 paths between `v4.0.6` and `v4.0.7` — container tooling, developer helpers, frontend test suites, backend migrations, and server configuration were never described by any list. Every release reported `store-required` for reasons unrelated to the shipped app, and the over-the-air path was effectively unreachable.

- `no-app-impact` now covers `_docker/**`, `_devextras/**`, `docker-compose*.yml`, `frontend/tests/**`, and the PHP tooling config.
- `backend-only` now covers `backend/migrations/**`, `backend/config/packages/**`, `backend/config/services.yaml`, and the `AI`, `DTO`, `Model`, and `Prompt` namespaces.
- `storeRequired` matches `*Mobile*`, `*Payment*`, `*Purchase*`, `*Subscription*`, and `*Stripe*` anywhere in the name, so `MobilePurchaseService.php` is gated by a named rule rather than by the fallback.
- The dependency rule is scoped to the root and `frontend/` manifests, which are the only ones that can reach the bundle.

Frontend application code and `cloudflare/**` stay fail-closed deliberately: the first is executable code inside the bundle, the second sits in the SSE request path.

Audited across all 1785 tracked files: no path is downgraded out of a named rule except the four dev-tool manifests under `_devextras/`. Everything else moved out of the fallback. The range still classifies as `store-required`, now for 16 real reasons instead of 96 phantom ones.


cristiangrxs:
Thanks — the underlying need is real, but the stated mechanism is not: the workflow dispatch API rejects a commit SHA as `ref` and answers `No ref found for: <sha>` ([github/docs#590](https://github.com/github/docs/issues/590)). `gh workflow run … --ref <sha>` therefore never worked, and requiring the default branch does not take that away.

The scenario itself is worth supporting, so the commit is now selected through an input instead:

- `commit` defaults to empty, which releases the tip of the default branch as before.
- A supplied value must be a full 40-character SHA, must exist, and must be an ancestor of the default branch, so an unreviewed commit is still refused.
- The workflow file and the classifier keep running from the default branch, because `ref` must still be that branch. The selected commit stays data rather than code, which preserves the trust assumption of the earlier `workflow_run` trigger.
- The existing green-commit check already keys off the same SHA and applies unchanged.

PR #1455, opened 2026-08-06
fix(widget): keep the embed host transparent so the page stays visible

### LINKED ISSUE #1450: Widget: host page obscured by opaque fullscreen overlay on open
## Problem
Opening the embedded chat widget paints an opaque fullscreen layer over the host page, so the underlying site appears to disappear while the chat panel remains visible.

---

## Expected
The host page stays fully visible behind/around the widget. Only the chat button and chat window (and, when enabled, the intentional fullscreen backdrop) may cover page content.

## Actual
On first open (lazy load), `#synaplan-widget-host` covers the entire viewport with a solid background. The host site is hidden under that layer; the chat UI still renders and can reply.

---

## Steps to reproduce
1. Embed the public widget script on an external site (e.g. a third-party landing page).
2. Leave the widget closed — the host page renders normally with only the floating button.
3. Click the widget button to open the chat.
4. Observe the host page content vanish into a solid dark/opaque full-viewport surface while the chat panel stays in the corner.

---

### Code
`SynaplanWidget.loadChat()` mounts a full-viewport host:

```ts
// frontend/src/widget.ts — container styles
position: fixed !important;
inset: 0 !important;
z-index: 2147483647 !important;
pointer-events: none !important;
```

The same method injects `style.css` into the Shadow DOM and remaps document selectors onto the host:

```ts
// frontend/src/widget.ts
const adaptedCss = widgetStyles.default
  .replace(/:root\b/g, ':host')
  .replace(/\bbody\b/g, ':host')
```

App base styles then apply to `:host`, including an opaque background:

```css
/* frontend/src/style.css — @layer base */
body {
  background: var(--bg-app);
}
```

After the rewrite this becomes `:host { background: var(--bg-app); }` on a `position: fixed; inset: 0` element, which visually covers the host document. `pointer-events: none` lets clicks pass through, so the failure is visual, not input-blocking.

### Observed console (unrelated to the overlay)
Realtime token requests fail with 404 while chat still works:

```
POST https://web.synaplan.com/api/v1/realtime/widget/<widgetId>/sessions/<sessionId>/token 404 (Not Found)
```

Those errors do not explain the blank host page; the overlay reproduces from the CSS host adaptation alone.

---

## Fix direction
Keep CSS custom properties on `:host`, but do not map document-level `html`/`body` rules (`background`, `overflow`, `height`, etc.) onto the full-viewport widget host. The host surface must remain transparent outside the chat chrome / explicit fullscreen backdrop.

---

## Verification
1. Embed the widget on an external page with visible content outside the chat corner.
2. Open the widget (lazy and eager paths).
3. Confirm host content remains visible and unobscured.
4. Confirm intentional fullscreen mode (when enabled) still shows its backdrop and can be closed.
5. Confirm light and dark widget themes both leave the host page visible.

---

## Notes
- related: Shadow DOM style adaptation in `frontend/src/widget.ts` (`loadChat`)
- Area: embedded widget host integration

## Screenshots

Image

Image

cristiangrxs:
## Summary
Opening the embedded chat widget painted an opaque full-viewport layer over the host page; the widget host now stays transparent so the embedding site remains visible.

## Changes
- `frontend/src/widget.ts`: stop rewriting `body` selectors to `:host` when injecting the app stylesheet into the Shadow DOM. The rewrite put `background: var(--bg-app)` (plus `height: 100%` / `overflow: hidden`) on `#synaplan-widget-host`, a `position: fixed; inset: 0` layer, which covered the whole page. `html` / `body` rules never match inside a shadow root, so leaving them alone is a no-op for the widget.
- New `frontend/src/utils/widgetShadowStyles.ts` with `adaptStylesForShadowDom()`: the adaptation is now a single documented, unit-tested function that only remaps `:root` → `:host` (the design tokens the widget UI reads).
- `frontend/src/widget.ts`: move `-webkit-font-smoothing` / `-moz-osx-font-smoothing` onto the shadow root element — previously they were inherited from the remapped `body` rule, everything else it contributed was already set explicitly on that element.
- Tests: unit tests for the CSS adaptation and an `@ci` e2e regression test asserting the host layer paints nothing.

## Verification
- [ ] Not tested (explain)
- [x] Manual
- [x] Tests added/updated

Manual check against the dev stack with the widget embedded on an external page (`localhost:3000` → API on `localhost:8000`), covering the issue's verification list:

| Scenario | Host `background-color` | Host page visible |
| --- | --- | --- |
| lazy, light theme | `rgba(0, 0, 0, 0)` | yes |
| lazy, dark theme | `rgba(0, 0, 0, 0)` | yes |
| eager, light theme | `rgba(0, 0, 0, 0)` | yes |
| eager, dark theme | `rgba(0, 0, 0, 0)` | yes |
| fullscreen mode | `rgba(0, 0, 0, 0)` | intentional backdrop `rgba(0, 0, 0, 0.4)` renders and closes the chat on click |

The new e2e test was confirmed to fail against a build with the old `body` → `:host` rewrite and to pass with the fix. Full frontend gate is green: `make -C frontend lint`, `npm run check:types`, `make -C frontend test` (1026 tests), plus `widget.spec.ts --grep "@ci"` (4 passed).

## Notes
Fixes #1450.

The 404s on `/api/v1/realtime/widget/.../token` seen in the issue's console output are unrelated to the overlay and are not addressed here.

Mobile impact: the fail-closed classifier marks `frontend/src/**/*.ts` as `store-required`. The change only affects the standalone embeddable widget bundle (`dist-widget`), which the mobile app does not ship.

## Screenshots/Logs
Before — host page hidden behind the opaque layer (light theme, chat open):

Before

After (local repro, chat open on a colored host page):

```
[lazy-light]  {"hostBackground":"rgba(0, 0, 0, 0)","heroVisible":true,"backdrop":null}
[lazy-dark]   {"hostBackground":"rgba(0, 0, 0, 0)","heroVisible":true,"backdrop":null}
[eager-light] {"hostBackground":"rgba(0, 0, 0, 0)","heroVisible":true,"backdrop":null}
[eager-dark]  {"hostBackground":"rgba(0, 0, 0, 0)","heroVisible":true,"backdrop":null}
[fullscreen]  {"hostBackground":"rgba(0, 0, 0, 0)","backdrop":"rgba(0, 0, 0, 0.4)","backdropClosesChat":true}
```

### DIFF
--- a/frontend/src/utils/widgetShadowStyles.ts
+++ b/frontend/src/utils/widgetShadowStyles.ts
@@ -1,0 +1,17 @@
+/**
+ * Adapts the application stylesheet for injection into the widget's Shadow DOM.
+ *
+ * Only `:root` is remapped: it carries the design tokens the widget UI reads
+ * (`--bg-app`, `--txt-primary`, …), and inside a shadow tree those have to live
+ * on `:host`.
+ *
+ * Document-level `html` / `body` rules are deliberately left untouched — they
+ * simply never match inside a shadow root. Remapping them onto `:host` painted
+ * the widget host, a `position: fixed; inset: 0` layer spanning the whole
+ * viewport, with the app background and hid the embedding page behind an opaque
+ * surface. The host must stay transparent; the chat chrome and the explicit
+ * fullscreen backdrop are the only surfaces allowed to cover host page content.
+ */
+export function adaptStylesForShadowDom(css: string): string {
+  return css.replace(/:root\b/g, ':host')
+}
--- a/frontend/src/widget.ts
+++ b/frontend/src/widget.ts
@@ -19,6 +19,7 @@
 
 import type { App } from 'vue'
 import { detectApiUrl } from './widget-utils'
+import { adaptStylesForShadowDom } from './utils/widgetShadowStyles'
 
 interface WidgetRealtimeConfig {
   /** Master kill-switch echoed by the backend. Defaults to true. */
@@ -429,12 +430,7 @@ class SynaplanWidget {
       // Inject styles into Shadow DOM (not document.head!)
       const styleEl = document.createElement('style')
       styleEl.id = 'synaplan-widget-styles'
-      // Adapt global selectors (:root, body) for Shadow DOM context
-      // In Shadow DOM, :root and body don't exist - use :host instead
-      const adaptedCss = widgetStyles.default
-        .replace(/:root\b/g, ':host')
-        .replace(/\bbody\b/g, ':host')
-      styleEl.textContent = adaptedCss
+      styleEl.textContent = adaptStylesForShadowDom(widgetStyles.default)
       shadow.appendChild(styleEl)
 
       // Inject markdown-specific styles for tables, code blocks, etc.
@@ -454,6 +450,8 @@ class SynaplanWidget {
         font-size: 16px;
         line-height: 1.5;
         box-sizing: border-box;
+        -webkit-font-smoothing: antialiased;
+        -moz-osx-font-smoothing: grayscale;
       `
           .replace(/\s+/g, ' ')
           .trim()
--- a/frontend/src/widget.ts
+++ b/frontend/src/widget.ts
@@ -83,6 +83,8 @@ interface WidgetConfig {
   poweredByUrl?: string
 }
 
+const DEFAULT_ICON_COLOR = '#ffffff'
+
 class SynaplanWidget {
   private config: WidgetConfig | null = null
   private button: HTMLElement | null = null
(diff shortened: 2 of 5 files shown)


cristiangrxs:
Follow-up review of this branch: the CSS change checks out. With a fresh `dist-widget` build I confirmed in Chromium that `#synaplan-widget-host` computes to `rgba(0, 0, 0, 0)` / `background-image: none` in light and dark theme and the embedding page stays fully visible, and that the dropped `html` / `body` declarations were no-ops for the host — `overflow: hidden` and `overscroll-behavior: none` never applied to the chat panel (it is `position: fixed`, so it escapes the host's clip and scrollport), and the compiled `:root` block contains custom properties only, which still land on `:host`.

One unrelated defect in the same file is fixed in 8deadc285:

- **XSS via the custom launcher icon.** `getIconContent()` interpolated `buttonIconUrl` into an `innerHTML` string, while `WidgetService::sanitizeConfig()` only prefix-checks that value (`http://`, `https://`, `data:image/`). A stored URL like `https://example.com/icon.png" onerror="…` closed the `src` attribute and ran script on every embedding page. The icon is now built with `document.createElement('img')` + `img.src`, so the whole value stays a URL. `frontend/tests/unit/widgetLauncher.spec.ts` covers it (verified it fails against the previous implementation).
- The loading spinner interpolated `this.config?.iconColor` without the `#ffffff` fallback the built-in icons use, so it could render `stroke="undefined"`. Both now use one `DEFAULT_ICON_COLOR` constant.

Gate green: `make -C frontend lint`, `npm run check:types`, `make -C frontend test` (1028 tests).

PR #1465, opened 2026-08-07
chore(frontend): add a tool for capturing App Store gallery screenshots

cristiangrxs:
## Why

The Umbrel App Store submission needs gallery screenshots, and there will be more of them (Elestio, and every listing after it). Taking them by hand produces a different crop and a different set of views every time, and the first attempt showed the real risk: the shots came out full of test-provider placeholder text and empty states, which says nothing about the product.

## What

`frontend/tests/e2e/capture-store-screenshots.js` — a hand-run tool, not part of the E2E suite:

```bash
SCREENSHOT_WIDGET_ID=wdg_… node tests/e2e/capture-store-screenshots.js
```

It drives the real UI rather than seeding content through the API, because the chat view opens on whichever conversation was last active — a seeded exchange is not necessarily the one that ends up in frame.

The captured views deliberately show work in progress instead of empty states:

| Shot | View | What it shows |
| ---- | ---- | ------------- |
| 1 | Chat | Answers grounded in a selected knowledge folder |
| 2 | `/files/search` | The query is actually executed, with real similarity scores |
| 3 | Chat widgets | Widget list with live session and message counters |
| 4 | Widget conversations | A visitor conversation opened, including the human-handoff action |
| 5 | Memories | The memory list |

Two details worth calling out:

- The sidebar version pill is hidden. It reads `dev` from a source build, while a packaged install shows a real version there — leaving it in a store gallery would misrepresent what a user gets.
- The success toast on the search view is waited out, so the frame shows results rather than a transient notification.

`WidgetSessionsView.vue` gains `data-testid="item-widget-session"` on the session row so the tool can open a conversation without matching on its text. That is the only change to shipped code.

## Verification

- `make -C frontend lint` — clean
- `docker compose exec -T frontend npm run check:types` — clean
- `make -C frontend test` — 118 files, 1028 tests passing
- Tool run end to end against the local stack; all five shots reviewed individually

Backend untouched, so the backend gate does not apply.

## Mobile classification

**backend-only** in effect — no bundled SPA behavior changes. The only shipped-code change is a test id on an element that renders identically before and after.

### DIFF
--- a/frontend/src/views/WidgetSessionsView.vue
+++ b/frontend/src/views/WidgetSessionsView.vue
@@ -305,6 +305,7 @@
                     ? 'bg-[var(--brand)]/10 shadow-sm'
                     : 'hover:bg-white/5 dark:hover:bg-white/5',
                 ]"
+                data-testid="item-widget-session"
                 @click="viewSession(session)"
               >
                 <div class="flex items-start gap-3">
--- a/frontend/tests/e2e/capture-store-screenshots.js
+++ b/frontend/tests/e2e/capture-store-screenshots.js
@@ -1,0 +1,188 @@
+/**
+ * Captures the App Store gallery screenshots.
+ *
+ * Not part of the E2E suite — a one-off tool, run by hand against a stack that
+ * holds presentable content: a knowledge folder with real documents, a widget
+ * with real sessions, and a chat model that is not the test provider.
+ *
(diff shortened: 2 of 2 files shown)


cristiangrxs:
Closing: the screenshot tooling is not wanted in the repository.

PR #1467, opened 2026-08-07
feat(release): drive the mobile app chain from published GitHub releases

cristiangrxs:
## Summary

- **The mobile chain now starts only when a GitHub release is published.** `mobile-release-artifacts.yml` runs on `release: published` (pre-releases are skipped), verifies a green CI run for the released commit, pulls the commit-matching `openapi-spec` artifact from that run, and classifies everything merged **since the previously published release** — one release, one classification, at most one delivery. Merges to `main` and tags alone never reach the apps anymore.
- **`release-tag.yml` no longer gates on the classification.** It always cuts the next patch tag on request (the platform jobs need tags regardless) and now prepares a **draft release with generated notes**. Publishing the draft is the deliberate release decision; the classification is still reported in the run summary as a preview.
- **The classification policy ships as little as possible through Apple.** The entire frontend web layer (`frontend/**`: Vue/TS code, styles, i18n, assets, generated schemas, web dependencies) is `ota-candidate` — permitted over the air by Apple ADPLA 3.3.2 for WebView-interpreted code. All server-side code and server-delivered plugins (`backend/**`, `plugins/**`) are `backend-only` and never trigger an app delivery. `store-required` is reserved for native projects, IAP/payments/entitlements, authentication transport, native seams, forced-update logic, and the update mechanism itself.
- **Every tracked path is now classified.** Previously 582 of 1931 tracked files hit the fail-closed fallback (which is why the last eight releases were all `store-required`); the audit now reports zero fallback hits. The fallback stays as the safety net for future unclassified paths, and `AGENTS.md` requires new paths to be classified in the same PR.

Companion docs PR in the apps repository: metadist/synaplan-apps — `docs/release-driven-mobile-chain`.

## Verification

- `node --test tests/*.test.mjs` — 37/37 green (classifier + next-release-tag).
- Full-repo audit `node scripts/mobile-impact.mjs --files <all tracked files>`: 0 fail-closed hits; only 41 genuinely sensitive files remain `store-required`.
- Simulation `--base v4.0.8 --head main`: 276 changed files → 5 store-required (subscription/paywall + auth/config stores), 28 ota, 145 backend-only, 98 no-impact.
- Both workflow files parse cleanly.

## Test plan

- [ ] After merge: `gh workflow run release-tag.yml -f dry_run=true`, then a real run — verify tag + draft release appear.
- [ ] Publish the draft and watch `mobile-release-artifacts.yml` classify against v4.0.8 and dispatch the sync.


cristiangrxs:
Thanks — checked this against the live API before changing anything.

`head_sha` **is** a documented and working query parameter of `GET /repos/{owner}/{repo}/actions/runs`; a request for `1d6889be…` returns only runs at that commit, so the workflow was not picking unrelated runs.

The underlying concern is still worth closing, though: GitHub ignores an *unknown* query parameter silently, so a typo or a future API change would quietly degrade this from "the CI run for this commit" to "the newest green CI run" without failing anything. Both release guards now re-check `head_sha` on the returned runs instead of trusting the server-side filter alone (`ca3ff85`):

```
gh api ".../actions/runs?head_sha=${RELEASE_SHA}&status=success&per_page=100" |
  jq -r --arg sha "$RELEASE_SHA" \
    '[.workflow_runs[] | select(.name == "CI" and .event == "push" and .head_sha == $sha)] | first | .id // empty'
```

Verified against the real API for `v4.0.15`: the run selection and the green-commit count both resolve correctly.

PR #1476, opened 2026-08-09
feat(gateway): configurable image handling and a full settings UI for AI Agents

orgaralf:
## Summary
Gives the AI Agents admin page real depth for tool calling, vision and image detail handling: every gateway setting is now configurable from the UI instead of only in `BCONFIG`, image handling gains its own transport controls, and the page now states what the gateway actually runs rather than only what it was configured to prefer.

The first commit is the follow-up work to #1475 that missed the merge window — it landed on that branch about two minutes after the squash-merge, so it never reached `main`. It is replayed here unchanged on top of current `main`.

## Changes

**New image controls** — `VISION_IMAGE_DETAIL` (`auto`/`low`/`high`, forwarded as OpenAI `image_url.detail`) and `VISION_MAX_IMAGES` (cap forwarded image blocks, oldest replaced by a placeholder). A new `VisionPolicy` applies them to the request body *before* the vision routing decision, so the router sees the images that actually reach the upstream. The two layers are independent: the mode decides who reads an image turn, detail and the cap decide what goes on the wire in every mode.

**Generalized settings endpoint** — `PUT /api/v1/messages-gateway/flags` accepts any subset of the gateway settings across three value kinds (boolean, enum, integer) with per-kind validation, and rejects a `synaplan` mode the install cannot honour (no search provider / no PIC2TEXT model) instead of silently storing it.

**A settings page that answers, not just asks** — the mode dropdowns read "automatic" a lot, and automatic resolves against things an operator cannot see from that page. `GatewayToolCatalog` now also reports the built-in tools it would run without a request in hand, reusing the same availability predicates as the request path so the two cannot drift, and the status endpoint exposes it as `server_tools`. The Tool calling section leads with that readout.

**Sectioned, modular panel** — the settings block moves out of `MessagesGatewayConfiguration.vue` into Access / Tool calling / Images / Context and session / Connection, each its own presentational component that emits the patch it changed. The repeated select and number controls become `GatewaySettingSelect` and `GatewaySettingNumber`, so clamping and the "nothing changed, save nothing" rule live in one place. Each control saves on change and reports the outcome; unavailable options stay visible but disabled with an inline reason, and a stored mode whose provider disappeared falls back to `auto`.

i18n for all four locales, and the gateway docs describe both image layers and the readout.

## Verification
- [x] Manual
- [x] Tests added/updated

Full gate green: `make -C backend lint`, `make -C backend phpstan` (0 errors), `make -C backend test` (3410 tests), `make -C frontend lint`, `vue-tsc`, `make -C frontend test` (1041 tests).

End-to-end, the status endpoint reports the readout as real data:

```
server_tools           = ["web_search", "analyze_image"]
web_search_available   = true
vision_available       = true
```

The demo changes each mode and watches the readout follow: switching web search to passthrough drops the `web_search` chip, switching it back restores it, and turning vision off drops `analyze_image`.

[ai_agents_settings_tool_and_image_controls.mp4](https://cursor.com/agents/bc-019fe7e2-36fa-7d59-831f-c5ebc52c3e1a/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fai_agents_settings_tool_and_image_controls.mp4)

Contrast was measured rather than eyeballed, since the readout is a new surface. Light 4.89:1 and dark 6.66:1 for the smallest text (the tool identifiers), everything else 4.68:1 or better — all above the 4.5:1 AA threshold. Checked in light, dark, V2 (the default) and V1.

[Tool calling readout, light theme](https://cursor.com/agents/bc-019fe7e2-36fa-7d59-831f-c5ebc52c3e1a/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fai_agents_tool_calling_readout_light.webp)

[Tool calling readout, dark theme](https://cursor.com/agents/bc-019fe7e2-36fa-7d59-831f-c5ebc52c3e1a/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fai_agents_tool_calling_readout_dark.webp)

## Notes
Supersedes the orphaned tail of #1475. Mobile impact classifies as `ota-candidate` (`node scripts/mobile-impact.mjs`), so no policy allow-list change was needed.

One unrelated observation while testing: `useDesignVariant()` is never called anywhere, and `design-v2` is hardcoded on `<html>` in `frontend/index.html`, so the V1/V2 switch in that composable has no effect. Left alone here.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-019fe7e2-36fa-7d59-831f-c5ebc52c3e1a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-019fe7e2-36fa-7d59-831f-c5ebc52c3e1a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/backend/src/AI/Messages/MessagesGateway.php
+++ b/backend/src/AI/Messages/MessagesGateway.php
@@ -8,6 +8,7 @@ use App\AI\Credential\UserProviderKeyResolver;
 use App\AI\Messages\Tools\GatewayToolCatalog;
 use App\AI\Messages\Tools\GatewayToolLoop;
 use App\AI\Messages\Translator\AnthropicPassthroughTranslator;
+use App\AI\Messages\Vision\VisionPolicy;
 use App\Entity\Model;
 use App\Entity\User;
 use App\Message\SummarizeApiSessionCommand;
@@ -27,7 +28,9 @@ use Symfony\Component\Messenger\MessageBusInterface;
  * credential resolve, optional body mutation, translator dispatch, metering.
  *
  * The server-side tool loop (MCP tools plus Synaplan's built-in web search) and
- * context injection are gated by config flags (default off).
+ * context injection are gated by config flags (default off). Image blocks pass
+ * through {@see VisionPolicy} first, which enforces the operator's vision
+ * settings and resolves the image detail the translators forward.
  *
  * @phpstan-type GatewaySuccess array{
  *     ok: true,
@@ -60,7 +63,13 @@ use Symfony\Component\Messenger\MessageBusInterface;
  *     }|null,
  *     replaced_server_tools: list<string>,
  *     web_search: string,
- *     vision: string,
+ *     vision: array{
+ *         handling: string,
+ *         mode: string,
+ *         detail: string,
+ *         images_forwarded: int,
+ *         images_omitted: int
+ *     },
  *     context_hash: string|null,
  *     debug: bool
  * }
@@ -100,6 +109,7 @@ final readonly class MessagesGateway
         private AnthropicPassthroughTranslator $anthropicPassthrough,
         private GatewayToolCatalog $toolCatalog,
         private GatewayToolLoop $toolLoop,
+        private VisionPolicy $visionPolicy,
         private MessagesContextInjector $contextInjector,
         private CacheItemPoolInterface $cache,
         private MessageBusInterface $messageBus,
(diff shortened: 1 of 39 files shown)


orgaralf:
Addressed Copilot's review: the MCP server summary now passes `{ count }` for interpolation alongside the plural choice, so 2+ servers render as e.g. \"2 connected tool servers\" instead of literal `{count}`.

PR #1479, opened 2026-08-10
fix(gateway): keep empty JSON Schema objects when re-encoding Claude Code tools

orgaralf:
## Summary
- Claude Code was failing against Synaplan with `400 tools.5.custom.input_schema.properties: Input should be an object` (and a follow-on invalid draft 2020-12 error on TaskCreate).
- Root cause: when the gateway mutates the request (injects `analyze_image` / web search) it drops `raw_body` and re-encodes. PHP turns empty JSON objects `{}` into `[]`, which Anthropic rejects.
- Fix: normalize tool `input_schema` object fields (`properties`, `additionalProperties`, …) back to empty objects before encoding on the Anthropic passthrough path.

## Test plan
- [x] Unit tests for empty `properties` and empty `additionalProperties`
- [x] Passthrough translator re-encode test
- [x] `make -C backend lint` / `phpstan` / full `make -C backend test`
- [x] Live Claude Code CLI against local Synaplan returns `synaplan-claude-ok`
- [ ] After merge + deploy: Claude Code against web.synaplan.com

### DIFF
--- a/backend/src/AI/Messages/AnthropicJsonSchemaNormalizer.php
+++ b/backend/src/AI/Messages/AnthropicJsonSchemaNormalizer.php
@@ -11,7 +11,8 @@ namespace App\AI\Messages;
  * `['properties' => []]`. A later {@see json_encode} then emits
  * `"properties":[]`, which Anthropic rejects (`input_schema.properties: Input
  * should be an object`). The same corruption hits empty schema objects used as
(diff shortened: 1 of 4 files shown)


orgaralf:
Addressed Copilot's review: empty nested schema objects (property schemas of \`{}\`, \`items: {}\`, and an empty root schema) are now restored to JSON objects as well, not only \`properties\` / \`additionalProperties\` maps.

PR #1483, opened 2026-08-10
fix(frontend): resolve App Store review findings for the native app

cristiangrxs:
App Review rejected iOS **4.0.0 (128)** on 10 August with three findings. All three live in this repository's SPA — no native code, no new plugins, no permission changes. The branch also carries five mobile UI defects reported against the app, which ship in the same binary.

## 2.3.10 — Google Play named in the iOS binary

`subscription.native.storeNote` hard-coded *"the App Store or Google Play"*, so the iOS build advertised a competing platform. The key is now split into `storeNoteIos` / `storeNoteAndroid` and `SubscriptionView` picks the right one via `getNativePlatform()`.

## 2.1(a) — Microphone showed an error instead of a permission prompt

The reviewer saw *"Microphone access denied. Please allow microphone access in your browser settings."*

Root cause: `useWebSpeech` only checked that `webkitSpeechRecognition` exists. iOS WKWebView exposes the constructor but has no recognition service behind it, so the request failed with `not-allowed` **before the system ever asked for the microphone** — a permission error for a permission that was never requested. The working path (record via `getUserMedia`, transcribe on the server) was therefore never reached.

- `useWebSpeech` is now `isWebSpeechSupported() && !isNativeApp()`.
- `showMicrophoneButton` follows the same rule, so the button only appears when a path that works is available.
- `AudioRecorder` no longer carries hard-coded English strings. Errors expose an i18n key, and the permission case routes to an iOS, Android, or web variant — no more "browser settings" inside an app that has no browser UI.

## 2.1 — No place to consent to AI processing

Apple asked where users consent to their input reaching third-party AI providers. There was no such place. The first onboarding screen now states, above the CTA, that messages, files, images and voice recordings are passed to AI providers to generate an answer, and that account information is not part of those requests. It links to the privacy policy. The existing "get started" button becomes the affirmative act — no extra screen, no extra hurdle.

Checking that screen in the simulator showed the disclosure was rendered in `txt-tertiary`, which is `txt-secondary` at 60% opacity: roughly 3.1:1 in light and 4.2:1 in dark, below the 4.5:1 AA floor. A notice the user consents to has to be readable, so it now uses `txt-secondary` in both themes.

## Reported UI defects

Five issues from `metadist/synaplan-apps`, all fixable in the web layer:

- **Biometric lock re-prompted after every attachment** (metadist/synaplan-apps#28). The system file/photo picker backgrounds the WebView, and `useBiometricLock` re-armed the lock on every `isActive: false` without qualification, so picking a file always cost a new prompt. Backgrounding still locks — the app switcher must not show the session — but a return within 60 seconds now lifts it silently. The window is longer than the 30 seconds `nativeLifecycle` uses for session re-checks because browsing a large photo library regularly takes longer than that. A lock the background event did not arm itself (cold start, or a prompt the user cancelled) is never eligible, so stepping out and back in is not a way around the lock.
- **Copy button floated in an empty bubble** (metadist/synaplan-apps#31). It rendered on the assistant role alone, while its handler already refused to copy nothing. Both now read the same `copyableText`. The button stays mounted through the stream so the bubble does not jump when the first token lands, and only disappears once a finished turn turns out to have no text.
- **Status bar overlapped the page** (metadist/synaplan-apps#33). The WebView runs edge-to-edge and the chat deliberately scrolls messages up behind the notch, which left them colliding with the clock. An opaque band inside the content layer covers `env(safe-area-inset-top)` in the same token the layer paints itself with, so there is no seam in either theme or in V2. It slides and clips with the drawer, and collapses to nothing wherever the inset is 0.
- **Delete button moved from card to card** (metadist/synaplan-apps#34). The download button could not shrink below its label — German "Herunterladen" on a two-column tile — so it pushed the icon actions out of place, by a different amount depending on which optional ones the file had. The label now yields and the icons hold their size.
- **Status badge ran under the action icons** (metadist/synaplan-apps#35). `FileVectorPill` and `FileSourceBadge` carried `whitespace-nowrap` on the outer element, which held them at full text width and defeated their own inner `truncate`. The nowrap moves to the label; the badges may shrink. Fixed in the components, so every use site is covered at once.

metadist/synaplan-apps#29 and metadist/synaplan-apps#32 are **not** addressed here: both need native changes and are deferred until iOS is approved.

## Scope

Native-only in effect: the onboarding flow runs exclusively in the app, and the store note sits behind `isNative`. Web and self-host behaviour is unchanged apart from the microphone error texts, which are now translated instead of English-only, and the UI fixes above, which are corrections in both places. The status-bar band is inert wherever the safe-area inset is 0, which includes every desktop browser.

## Tests

Five new specs pin the findings so they cannot come back:

- `SubscriptionViewStoreNote.spec.ts` — iOS never renders "Google Play", Android never renders "App Store", checked across all four locales.
- `OnboardingWelcomeStep.spec.ts` — the notice is present with a working privacy link in every locale.
- `useBiometricLock.spec.ts` — cold start prompts, a short excursion does not, a real absence does, and a cancelled unlock stays locked. The file had no tests before.
- `ChatMessageCopyButton.spec.ts` — the button follows the copyable text, survives an empty stream, and never appears on a user message.
- `FileBadgeTruncation.spec.ts` — the badges can shrink while their labels stay on one line.

Gate: `make -C frontend lint`, `make -C frontend test` (125 files, 1073 tests) green. `vue-tsc` shows the same 14 pre-existing errors in `messagesGatewayApi.ts` on this branch as on a clean `main`; nothing here adds to them.

### Verified in the iOS simulator

Built from this branch and run on an iPhone 17 Pro Max (iOS 26.5) — the device Apple reviewed on — against a local backend:

- Microphone: tapping it starts a recording and returns a transcript. No error message.
- Subscription page: "purchases and subscriptions are handled through the **App Store**", renewal terms, both legal links and restore, all localised.
- Onboarding: the AI disclosure and its privacy link are on the first screen, on a clean install.
- Status bar: messages clip at the inset instead of colliding with the clock, in light and dark.
- Copy button: present on answers with text, absent on empty bubbles.
- Files: badges truncate inside their row, and the delete button holds the same position across cards with different actions.
- Biometric lock: a six-second excursion returns without a prompt; after 70 seconds the lock screen and Face ID prompt come back.

Android was built and installed from the same branch, but the emulator on this machine has no working network route, so anything behind the API could not be exercised there. What is reachable without a backend — launch, status-bar band, the server dialog and its URL validation — behaves correctly.

## Release classification

**store-required** — Apple explicitly asks for a revised binary, and `scripts/mobile-impact.mjs` classifies the branch the same way off `useBiometricLock.ts` and `SubscriptionView.vue`. Follow-up: pin sync and a new build in `synaplan-apps`.

### DIFF
--- a/backend/src/AI/Service/AiProviderDisclosure.php
+++ b/backend/src/AI/Service/AiProviderDisclosure.php
@@ -1,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\AI\Service;
+
+use App\AI\Credential\ChatReadinessService;
+use App\Repository\ModelRepository;
+
+/**
+ * MOBILE-APP SEAM (App Review 5.1.2(i) / Play "prominent disclosure"): names
+ * the AI providers a user's input can reach.
+ *
+ * Since November 2025 Apple requires the consent screen to identify the
+ * third-party AI by name — "external AI providers" is not enough. The app shows
+ * that screen before sign-in, so the list has to come from the public runtime
+ * config, and it has to reflect the instance the app is pointed at rather than
+ * a list baked into the bundle.
+ *
+ * Two filters make the list match reality rather than the catalog: the model
+ * has to be pickable in chat, and the provider has to be usable. A provider
+ * without a configured key cannot receive anything, and listing it would pad a
+ * disclosure that people only read while it stays short.
+ */
+final readonly class AiProviderDisclosure
+{
+    public function __construct(
+        private ModelRepository $modelRepository,
+        private ProviderRegistry $providerRegistry,
+        private ChatReadinessService $chatReadiness,
+    ) {
+    }
+
+    /**
+     * Display names of the providers behind the usable, selectable chat models.
+     *
+     * @return list<string> Sorted and deduplicated, e.g. ['Anthropic', 'Google AI']
+     */
+    public function chatProviderNames(): array
+    {
+        $displayNames = [];
+        foreach ($this->providerRegistry->getUniqueProviders() as $provider) {
+            $displayNames[strtolower($provider->getName())] = $provider->getDisplayName();
+        }
+
+        // Cached for 30s inside the service — the public config is polled, and
+        // a fresh probe per request would put an Ollama round trip on an
+        // unauthenticated endpoint.
+        $available = $this->chatReadiness->providerAvailability();
+
+        $names = [];
+        foreach ($this->modelRepository->findSelectableChatServices() as $service) {
+            $key = strtolower(trim($service));
+
+            // Unknown to the registry means there is no client that could send
+            // anything, so it belongs in no disclosure. 'test' is a fixture.
+            if ('' === $key || 'test' === $key || !isset($displayNames[$key])) {
+                continue;
+            }
+
+            if (!($available[$key] ?? false)) {
+                continue;
+            }
+
+            $names[$key] = $displayNames[$key];
+        }
+
+        $names = array_values($names);
+        sort($names, SORT_NATURAL | SORT_FLAG_CASE);
+
+        return $names;
+    }
+}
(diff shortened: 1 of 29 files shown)


cristiangrxs:
Copilot's point on the locale loop was right: it only covered iOS, so a translated Android string could have reintroduced the rejected wording without failing. `d93295a9b` now iterates both platforms across all four locales and asserts the expected store is present and the other absent.

PR #1512, opened 2026-08-19
feat(models): detect discontinued provider models before users hit them

FExB17:
## Why

Providers retire models without telling us, and so far we learned about it from a user-facing provider error. #1511 retires one such model by hand — this PR is the part that finds the next one first.

## What it does

`app:models:check-availability` compares what we offer against what providers actually serve, in two independent scopes:

- **`database`** — the active `BMODELS` rows of *this* install. `ModelSeeder` never deactivates anything, so a row the catalog dropped long ago can still be selectable here, and only the database knows that.
- **`catalog`** — what `ModelCatalog` still ships to *new* installs. An operator who cleaned up locally would otherwise never learn that every fresh install keeps getting the dead model.

## The part that makes it trustworthy

Detection is two-stage: the provider's model list is only a cheap pre-filter, and every model missing from that list is then confirmed individually via `GET {listUrl}/{modelId}`. Only a model the provider itself answers `404`/`400` for is reported.

A listing-only verdict is wrong in **both** directions — verified against the live APIs while building this:

| Case | Listing | Direct probe | A list-only check would… |
| --- | --- | --- | --- |
| `google imagen-4.0-generate-001` | absent | **200 — alive** (served via `:predict`) | raise a false alarm on all three Imagen rows |
| `xai grok-tts` | absent | **404 — gone** | look identical to the case above |

I first tried to separate the two with a per-capability coverage heuristic. It excused Imagen correctly and hid `grok-tts` — which is why the heuristic is gone and the probe decides.

## Deliberately read-only

No `BACTIVE = 0`, no default repointing. A model also vanishes from a listing through a rename, a region gate or an account tier, and an unattended deactivation on a false positive takes a working model away from every user of an install. Retirement stays a reviewed migration.

Nothing is ever read as evidence when it isn't:

- No key, no listing endpoint (HuggingFace validates via `whoami-v2`; Ollama et al. are per-install) or an unreachable API → **unchecked**, never "serves nothing". A conclusive-looking empty list is rejected too.
- A probe answering `401`/`429`/`5xx` → **unconfirmed**: visible in the console, excluded from the alert and the exit code.
- Probes are capped per provider so a malformed listing cannot fan out into hundreds of requests.

Findings on a `ProviderDefaultsService` recommendation are marked and sorted first, because `app:provider:apply-defaults --auto` assigns those unattended at container start.

## First run against the live APIs

Six confirmed retirements, each re-verified by hand with `curl` plus a known-good control:

| Provider | Model | BID | Note |
| --- | --- | --- | --- |
| groq | `meta-llama/llama-4-scout-17b-16e-instruct` | 17 | Groq `PIC2TEXT` default |
| groq | `llama-3.3-70b-versatile` | 9 | already handled by #1511 |
| groq | `qwen/qwen3-32b` | 53 | |
| groq | `llama-3.1-8b-instant` | 236 | |
| xai | `grok-stt` | 321 | xAI `SOUND2TEXT` default |
| xai | `grok-tts` | 320 | |

Two of them are provider defaults, which is exactly the class of breakage nobody chooses to run into. Retirement migrations are follow-ups, one per provider — this PR intentionally changes no model row.

## How it runs

Daily in the scheduler role with `--notify` (Discord, no-op without `DISCORD_WEBHOOK_URL`). Exit codes match `app:sync-model-prices`: `0` clean, `1` broke, `2` confirmed findings with `--fail-on-drift`. Installs without cloud provider keys make no outbound request at all.

## Test plan

- [x] `make -C backend test` — 3627 tests green
- [x] `bash _docker/backend/tests/test-container-runtime.sh` — 65/65, incl. the new scheduler assertion
- [x] Ran against live Groq/OpenAI/Anthropic/Google/Mistral/xAI APIs; every finding and every suppression re-verified by hand
- [x] `scripts/mobile-impact.mjs` → `backend-only`
- [ ] Reviewer: confirm the read-only stance is the wanted default rather than auto-deactivation

### DIFF
--- a/_docker/backend/lib/container-runtime.sh
+++ b/_docker/backend/lib/container-runtime.sh
@@ -352,7 +352,7 @@ run_scheduler_role() {
     mkdir -p "$SYNAPLAN_RUNTIME_DIR"
     trap stop_scheduler TERM INT
 
-    runtime_log "Starting scheduler (media reaper every ${tick_seconds}s, ephemeral-file reaper every ${hourly_seconds}s, update check every ${daily_seconds}s)."
+    runtime_log "Starting scheduler (media reaper every ${tick_seconds}s, ephemeral-file reaper every ${hourly_seconds}s, update + model-availability check every ${daily_seconds}s)."
     while [ "$_scheduler_stopping" -eq 0 ]; do
         now="$(date +%s)"
         printf '%s\n' "$now" > "${SYNAPLAN_RUNTIME_DIR}/scheduler.heartbeat"
@@ -380,6 +380,15 @@ run_scheduler_role() {
             if ! run_scheduler_command bin/console --env="$env" app:updates:check --no-interaction; then
                 runtime_log "Update check failed; it will be retried on the next daily interval." >&2
             fi
+
+            # Discontinued-model detection. Read-only and reports only: it asks
+            # the providers the operator already configured a key for which
+            # models they still serve, and never changes a row. Installs without
+            # cloud keys make no outbound request at all.
+            if ! run_scheduler_command bin/console --env="$env" app:models:check-availability --notify --no-interaction; then
+                runtime_log "Model availability check failed; it will be retried on the next daily interval." >&2
+            fi
+
             next_daily=$((now + daily_seconds))
         fi
 
--- a/_docker/backend/tests/test-container-runtime.sh
+++ b/_docker/backend/tests/test-container-runtime.sh
@@ -105,6 +105,7 @@ echo "Case 2: scheduler runs every slot's command initially and writes a heartbe
 assert_contains "app:media:reap-jobs --no-interaction" "$COMMAND_LOG" "scheduler runs media reaper"
 assert_contains "app:files:reap-ephemeral --no-interaction" "$COMMAND_LOG" "scheduler runs ephemeral-file reaper"
 assert_contains "app:updates:check --no-interaction" "$COMMAND_LOG" "scheduler runs the daily update check"
+assert_contains "app:models:check-availability --notify --no-interaction" "$COMMAND_LOG" "scheduler runs the daily model availability check"
 if [ -s "$TMP_DIR/runtime/scheduler.heartbeat" ]; then
     assert_eq 1 1 "scheduler writes liveness heartbeat"
 else
--- a/backend/config/services.yaml
+++ b/backend/config/services.yaml
@@ -582,6 +582,11 @@ services:
     App\Service\MediaGenerationServiceInterface:
         alias: App\Service\MediaGenerationService
 
+    # Model availability detection talks to provider listing endpoints; the
+    # interface keeps the diff logic testable without credentials or network.
+    App\AI\Service\ProviderModelInventoryInterface:
+        alias: App\AI\Service\ProviderModelInventory
+
     App\Service\Media\GeneratedFileRegistrar:
         arguments:
             $uploadDir: '%kernel.project_dir%/var/uploads'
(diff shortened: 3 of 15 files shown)


FExB17:
Merge scope of this PR: read-only detection and reporting. It shares no file with #1513 and merges cleanly in either order; the full backend gate is green on the merged state of both branches.

Findings from the first detection run are tracked separately and are not addressed here:

- #1514 — `xai:grok-stt` and `xai:grok-tts` confirmed retired upstream (`404` on the per-model endpoint, `200` for a control model), both still `'active' => 1` in `ModelCatalog`, and `grok-stt` still the recommended xAI `SOUND2TEXT` default.
- #1515 — the four Groq rows from the same run are retired by #1513; the underlying gap is that a model id is pinned in six places with no lifecycle state, so each retirement is a hand-written migration that heals only part of them.

cristiangrxs (approved):
Approving. I reproduced your findings independently, and the two-stage argument holds up under a test I did not expect to fail.

## Independent confirmation

I have been building a model health monitor on a separate branch (traffic telemetry, persisted state, alerting, admin status page). It shares nothing with this PR except the idea of reading provider catalogs, and it took the listing-only shortcut you warn about — including the exact per-capability coverage heuristic you say you tried and removed.

Your PR made me measure it. On our own catalog, with this install's keys, the heuristic was wrong in both directions:

| Model | Direct probe | Listing-only verdict |
| --- | --- | --- |
| `imagen-4.0-generate-001` | 200 — alive | reported retired |
| `imagen-4.0-fast-generate-001` | 200 — alive | reported retired |
| `imagen-4.0-ultra-generate-001` | 200 — alive | reported retired |
| `grok-tts` | 404 — gone | reported healthy |
| `grok-stt` | 404 — gone | reported healthy |

The reason is structural, and worth recording because it will tempt the next person too: the heuristic excuses a missing model when no other model of the same capability is listed. Google lists `gemini-2.5-flash-image` under `text2pic`, so the three Imagen rows lose their excuse and get condemned. xAI has exactly one `text2sound` row, so the genuinely dead `grok-tts` keeps its excuse forever. It tracks how our catalog is grouped, not what the provider serves. Your `GET {listUrl}/{modelId}` oracle is the only thing that separates the two cases.

After replacing the heuristic with a direct per-model probe, my monitor independently reports the same six retirements as this PR — the four Groq models and both xAI ones — from a completely separate code path.

## On your open checkbox

Read-only is the right default, and I would keep it even though my branch does have an auto-disable path. Mine is off by default and gated on confirmed evidence, and after the measurement above I am much less inclined to loosen that. A false positive here does not annoy an operator, it removes a working model from every user of an install. Retirement as a reviewed migration is the correct blast radius.

## One thing worth changing

`offeredModels()` keys on `providerId|tag`, so a model bound to several capabilities becomes several entries — and `run()` probes per entry. The findings are deduplicated afterwards, so the report is right, but the requests are not: a dead `claude-opus-5` (chat + pic2text) is probed twice at the same URL, and `claude-sonnet-5` three times. It also distorts `MAX_PROBES_PER_PROVIDER`, since duplicates spend budget — for Anthropic, where nearly every chat model carries a `pic2text` row too, the cap effectively halves to ~13 distinct models.

Deduplicating by provider model id before probing fixes both. I hit the same thing and now resolve presence once per provider model id, then apply it to every row that shares it.

Not blocking — the verdicts are correct either way.

## What I would like to keep from this PR

The `catalog` scope and the `ProviderDefaultsService` marking have no equivalent in my branch and should not be lost in any later consolidation. An install that cleaned up locally while every fresh install keeps seeding the dead model is a failure mode nothing else here would catch.

Once this lands I will rebase my branch on top. `ProviderModelInventory::probe()` and my probe layer are the same function against the same endpoints, so mine should use your `ProviderModelInventoryInterface` as the transport rather than keeping a second implementation. Our scheduler slots also need to become one run instead of two that poll the same endpoints.

PR #1519, opened 2026-08-19
fix: auto-redirect expired OIDC sessions back to the IdP

dschmidt:
On instances with `auto_redirect` enabled, an expired session landed users on the manual "Login with SSO" page even though a live SSO session at the IdP would silently re-log them in; OIDC-only deployments never want to see that page.

The session-expired path now redirects straight to the IdP as well, guarded by a per-browser-session rate limit (30s via sessionStorage): if the user bounces back to /login within the window, app-side token validation is failing while the IdP session is alive (e.g. a misconfigured audience mapper), and the existing manual sign-in page shows instead of looping. Without sessionStorage the guard fails safe to the manual page.

Verified: eslint, prettier, vue-tsc clean.


orgaralf (approved):
Thanks

PR #1522, opened 2026-08-20
fix(config): container-safe defaults for strictly-bound env vars

dschmidt:
## Problem

Bare images (Kubernetes via synaplan-charts, plain `docker run`) ship no `backend/.env`, so every `%env()%` binding without a parameter default crashes service instantiation as soon as the service is needed. Concretely: every chat message fails with **HTTP 500** because `WhatsAppService` strictly binds `WHATSAPP_ENABLED`:

```
Uncaught PHP Exception InvalidArgumentException: "The controller for URI \"/api/v1/messages/stream\" is not callable: Environment variable not found: \"WHATSAPP_ENABLED\"."
```

docker-compose deployments never see this because compose passes the vars through (`WHATSAPP_ENABLED: ${WHATSAPP_ENABLED:-false}`) and loads `backend/.env` via `env_file`.

## Fix

Add `env()` parameter defaults in `services.yaml` (the repo already does this for ~69 other vars) for the 20 strictly-bound vars that had none, values mirroring `backend/.env.example`:

- `WHATSAPP_ENABLED` (false)
- `TIKA_TIMEOUT_MS` / `TIKA_RETRIES` / `TIKA_RETRY_BACKOFF_MS` / `TIKA_MIN_LENGTH` / `TIKA_MIN_ENTROPY`
- `RASTERIZE_DPI` / `RASTERIZE_PAGE_CAP` / `RASTERIZE_TIMEOUT_MS`
- `WHISPER_ENABLED` / `WHISPER_DEFAULT_MODEL` / `WHISPER_BINARY` / `WHISPER_MODELS_PATH`
- `FFMPEG_BINARY`
- `BRAVE_SEARCH_ENABLED` / `BRAVE_SEARCH_API_URL` / `BRAVE_SEARCH_COUNT` / `BRAVE_SEARCH_COUNTRY` / `BRAVE_SEARCH_SEARCH_LANG`
- `STRIPE_AUTOMATIC_TAX`

`DATABASE_READ_URL` / `DATABASE_WRITE_URL` intentionally stay strict (fail fast without a database).

Audit method: diffed all `%env()%` bindings in `backend/config` + `backend/src` (excluding `default::`-processed ones) against existing `env()` parameter defaults; the listed 20 plus the two database URLs were the only strict leftovers.

`lint:container` passes with the change.

### DIFF
--- a/backend/config/services.yaml
+++ b/backend/config/services.yaml
@@ -122,9 +122,35 @@ parameters:
     env(IAP_STORE_PRICE_BUSINESS): '129.99'
     env(WHATSAPP_ACCESS_TOKEN): ''
     env(WHATSAPP_WEBHOOK_VERIFY_TOKEN): ''
+    env(WHATSAPP_ENABLED): 'false'
     env(CLOUDFLARE_ACCOUNT_ID): ''
     env(CLOUDFLARE_API_TOKEN): ''
     env(EMBEDDING_FALLBACK_PROVIDER): ''
+
+    # Container-safe defaults for the remaining strictly-bound env vars.
+    # Compose passes these through .env/env_file, but bare images (Kubernetes,
+    # plain docker run) ship no .env, so a missing var used to crash service
+    # instantiation (e.g. /api/v1/messages/stream 500ing on WHATSAPP_ENABLED).
+    # Values mirror backend/.env.example.
+    env(TIKA_TIMEOUT_MS): '30000'
+    env(TIKA_RETRIES): '2'
+    env(TIKA_RETRY_BACKOFF_MS): '1000'
+    env(TIKA_MIN_LENGTH): '10'
+    env(TIKA_MIN_ENTROPY): '3.0'
+    env(RASTERIZE_DPI): '150'
+    env(RASTERIZE_PAGE_CAP): '10'
+    env(RASTERIZE_TIMEOUT_MS): '30000'
+    env(WHISPER_ENABLED): 'true'
+    env(WHISPER_DEFAULT_MODEL): 'base'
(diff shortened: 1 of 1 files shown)


dschmidt:
Note on the seemingly inconsistent pair `env(WHATSAPP_ENABLED): 'false'` vs `env(BRAVE_SEARCH_ENABLED): 'true'`: these deliberately mirror `backend/.env.example` and the compose defaults (`WHATSAPP_ENABLED: ${WHATSAPP_ENABLED:-false}`). The whole point of this PR is that a bare image behaves exactly like a compose deployment, so the parameter defaults must not diverge from those, even where the existing choices look asymmetric.

The asymmetry also has a rationale: Brave is double-gated (flag + `BRAVE_SEARCH_API_KEY`, so `true` without a key is inert), while WhatsApp wires up an externally reachable webhook endpoint and stays opt-in.

If we want the two flags actually consistent (e.g. both `false` as secure default), that should be a separate change touching `.env.example`, compose and these parameter defaults together, not something this PR sneaks in.

orgaralf (approved):
Nice clean-up, thx!

dschmidt:
Correction: 13 of the 20 vars listed here (`BRAVE_SEARCH_*`, `TIKA_*`, `RASTERIZE_*`) were already baked as image `ENV` in the Dockerfile, so only `WHATSAPP_ENABLED`, `WHISPER_*`, `FFMPEG_BINARY` and `STRIPE_AUTOMATIC_TAX` were actually missing in bare images. Harmless (image env wins over parameter defaults), but `BRAVE_SEARCH_ENABLED` got `'true'` here while the image bakes `false`; aligned in #1583.

PR #1583, opened 2026-08-26
fix(config): default BRAVE_SEARCH_ENABLED to false like the image does

dschmidt:
Follow-up to #1522.

The Dockerfile already bakes `BRAVE_SEARCH_ENABLED=false` together with the `BRAVE_SEARCH_*`, `TIKA_*` and `RASTERIZE_*` values as image `ENV`, and a real env var wins over a `env()` parameter default. So the `'true'` default from #1522 only ever applied outside containers, and the flag behaved differently in-container vs. out. Align the parameter default with the image and note the precedence in the comment.

Correction to the #1522 description: of the 20 listed vars, 13 (`BRAVE_SEARCH_*`, `TIKA_*`, `RASTERIZE_*`) were already covered by the Dockerfile `ENV`; the ones that were actually missing in bare images were `WHATSAPP_ENABLED`, `WHISPER_*`, `FFMPEG_BINARY` and `STRIPE_AUTOMATIC_TAX`. The defaults are harmless either way (image env takes precedence), the count was just overstated.

### DIFF
--- a/backend/config/services.yaml
+++ b/backend/config/services.yaml
@@ -131,7 +131,9 @@ parameters:
     # Compose passes these through .env/env_file, but bare images (Kubernetes,
     # plain docker run) ship no .env, so a missing var used to crash service
     # instantiation (e.g. /api/v1/messages/stream 500ing on WHATSAPP_ENABLED).
-    # Values mirror backend/.env.example.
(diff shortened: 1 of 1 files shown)


orgaralf (approved):
Thanks for the cleanup 1!

PR #1584, opened 2026-08-26
feat(speech): WEB_SPEECH_ENABLED switch for the browser speech-to-text path

dschmidt:
## Problem

`ChatInput` picks the browser's Web Speech API whenever `SpeechRecognition` exists and only falls back to the record-then-transcribe path (local Whisper / API model) when it does not. Two things go wrong with that:

- The Web Speech API streams the microphone audio to the browser vendor's cloud recognizer (Google for Chrome). Air-gapped or data-residency-bound deployments cannot allow that, and there is no way to opt out.
- Chromium builds without Google API keys expose `webkitSpeechRecognition` but never deliver a result: the user sees "Listening... speak now" and nothing else. Feature detection cannot tell that apart from a working Chrome.

(The `toggleRecording` docstring claimed the choice honoured `whisperEnabled`; the code never did.)

## Change

- `WEB_SPEECH_ENABLED` env flag (default `true`, parsed like `REGISTRATION_ENABLED` / `GUEST_CHAT_ENABLED` via a small `WebSpeechConfig`), exposed as `speech.webSpeechEnabled` in `/api/v1/config/runtime`.
- Frontend: `useWebSpeech` additionally requires the flag; `showMicrophoneButton` now shows the mic when Web Speech is usable **or** the server can transcribe, so a deployment with the flag off and no Whisper/API model hides the mic instead of offering a dead path.
- Docs (`CONFIGURATION.md`, `.env.example`), compose passthrough (`deploy/compose.yaml`, `selfhost.env.example`).
- Tests: `WebSpeechConfigTest` (env parsing), existing `ConfigController` unit tests get the new stub.

No behaviour change for installs that do not set the flag.


orgaralf (approved):
Thanks for the cleanup 2!

PR #1585, opened 2026-08-26
fix(whisper): honour WHISPER_DEFAULT_MODEL for chat transcription

dschmidt:
`MessagePreProcessor::transcribeWithWhisper()` hardcoded `$options['model'] = 'base'`. `WhisperService::isAvailable()` checks the *configured* default model (`WHISPER_DEFAULT_MODEL`), so a deployment shipping e.g. `ggml-small.bin` only reported speech-to-text as available and then failed every chat transcription asking for `ggml-base.bin`.

Drop the override; `WhisperService::transcribe()` already falls back to its configured default when no model option is given. Behaviour is unchanged for installs that keep the default (`base`).

Found while wiring local Whisper on an air-gapped Kubernetes install where `small` is the sensible model for German.

### DIFF
--- a/backend/src/Service/Message/MessagePreProcessor.php
+++ b/backend/src/Service/Message/MessagePreProcessor.php
@@ -573,9 +573,6 @@ final readonly class MessagePreProcessor
                 $options['language'] = $languageHint;
             }
 
-            // Use base model by default (good balance of speed/accuracy)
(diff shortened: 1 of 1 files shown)


orgaralf (approved):
Thanks for the cleanup 3!

Issue #1586, opened 2026-08-26
Model catalog must reflect the providers of this installation


orgaralf:
## Context

Every install seeds the full model catalog (currently 117 models across 14 providers) as active + selectable — regardless of which providers the installation can actually reach. On installs with few or no cloud keys (on-prem, air-gapped, local-first hosters) users pick from dozens of models that cannot answer.

The availability truth already exists (`ChatReadinessService::providerAvailability()`, `ModelConfigService::usableProviders()`) and the routing path uses it — the catalog surfaces do not.

Related: #462 (on-prem tracker — this aspect was never part of it), #1521 (no-provider tombstone), #1516 (model health), #1532 (retirement data model).

**Example install (air-gapped):** Ollama with 3 pulled models, no cloud keys, `WEB_SPEECH_ENABLED=false`. Users should see exactly those models — today they see the whole catalog.

## Acceptance criteria

1. Provider availability is the backend truth already computed today (credentials/URL present; Ollama additionally requires the model to be pulled) — consumed, not reimplemented, and never per-model handwork.
2. `GET /api/v1/config/models`, all model pickers, and "Select suggested models" only show/recommend models of available providers. Regular users: unavailable models are **hidden**. Admin views: full catalog with unavailable entries **greyed out + badge** ("no key configured" for keyed providers, "not pulled" for Ollama models) linking to the provider setup page. Seeding stays provider-agnostic; filtering happens at **read time**, so saving a key in the admin UI surfaces its models without a re-deploy.
3. New users get no frozen per-user default model rows; global defaults apply until a user explicitly chooses. Local whisper.cpp becomes a catalog entry (service `Whisper`) so SOUND2TEXT can bind to it — air-gapped STT on every browser.
4. CLI: `app:provider:list`, `app:model:enable|disable --provider <name>`; disable is a **soft-deactivate** that survives `app:seed` (today's disable deletes the row and the next container start resurrects it). Catalog rows are never deleted — disabled/unavailable entries reappear as soon as their provider has credentials.
5. Helm chart/compose expose provider-level enable/disable consistent with the CLI, plus an air-gap example values file; documentation updated (admin guide + air-gapped installation guide).

## Plan

Full concept and sequencing in `_devextras/planning/20260826-provider-aware-model-catalog.md`: (A) CLI provider dimension + soft-disable fix (+ chart values), (B) availability-filtered read surfaces, (C) defaults un-freeze, (D) local Whisper catalog entry, (E) documentation.

[...]

PR #1587, opened 2026-08-26
feat(cli): provider dimension for model commands and soft-disable

orgaralf:
Part of #1586 (PR A of the plan).

## Summary
- `app:model:disable` no longer DELETEs catalog rows. Deleting broke `BMESSAGES` references and was silently reverted by the seeder re-inserting the row on the next container start. It now soft-disables (`BACTIVE=0`, `BSELECTABLE=0`) — operator-owned flags the re-seed preserves — and a row missing from the DB is inserted first so the deactivation sticks. Models reappear when re-enabled; nothing is ever deleted.
- `app:model:enable` restores the catalog visibility flags on existing rows without touching admin price/name edits, and skips retired models with a note.
- Both commands accept repeatable `--provider <name>` to act on a provider's complete catalog (`synaplan model:disable --provider openai`). Explicit model keys and provider selections are deduplicated, so per-model entries can refine provider-wide toggles.
- New `app:provider:list`: one table with each provider's credential source (`key (ui)` / `key (env)` / `none` / `n/a`), availability (reusing `ChatReadinessService`, incl. the Ollama pulled-model gate) and active/total model counts. `--fresh` bypasses the cached snapshot.

Companion chart PR: metadist/synaplan-charts#40 (provider-level values).

## Test plan
- [x] `make -C backend test` — 4104 tests green
- [x] New unit tests: `ModelCatalog::enable/disable/findByService/serviceNames`, soft-disable never issues DELETE, missing rows are inserted deactivated
- [x] Command tests: `--provider`, dedupe, retired-model skip, unknown provider/key failures, no-args rejection
- [x] `app:provider:list` smoke-tested against a live install

### DIFF
--- a/_devextras/planning/20260826-provider-aware-model-catalog.md
+++ b/_devextras/planning/20260826-provider-aware-model-catalog.md
@@ -1,0 +1,369 @@
+# Provider-Aware Model Catalog — Concept & Development Plan
+
+**Date:** 2026-08-26
+**Status:** AGREED — all open questions resolved, implementation in progress
+in this workspace (all PRs, no ownership split).
+
+**Decisions (2026-08-26):**
+- **Never delete rows from `BMODELS`.** Disabling — by operator, CLI, or
+  missing provider credentials — only greys out or hides entries. When a key
+  is added later, the entries reappear automatically (read-time filtering).
+  The CLI gets no deletion path at all; `ModelCatalog::remove` stays reserved
+  for the retirement machinery, which also deactivates rather than deletes.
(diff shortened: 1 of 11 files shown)


orgaralf:
Follow-up commit: \`app:model:enable --only <provider>\` (repeatable). Air-gap allow-list — enables the named providers and soft-disables every other catalog provider, so a new cloud model added upstream does not need a deny-list update. Chart companion: metadist/synaplan-charts#40 (\`models.providers.only\`).

PR #1588, opened 2026-08-26
feat(models): hide unavailable models from pickers, grey them for admins

orgaralf:
Part of #1586 (PR B of the plan).

## Summary
- `GET /api/v1/config/models` only returns models whose provider is available on this installation (API key / URL configured; Ollama models must be pulled). Regular users never see the rest.
- Admins can pass `includeUnavailable=1` to receive the full catalog, each row flagged with `available` / `unavailableReason` (`provider_unavailable` or `not_pulled`).
- The admin catalog (`GET /api/v1/admin/models`) always returns every row with `providerAvailable` + `unavailableReason` so the UI can grey unavailable entries and badge them.
- The response includes a `providers` block (`name`, `displayName`, `available`, `requiresKey`) so the frontend can hide **Select suggested models** unless every key-based cloud provider has a key. URL/local providers (Ollama, custom endpoints) do not count.
- Chat picker and settings dropdowns inherit the filtered list. Admin Edit Models greys rows and shows "Provider not configured" / "Not pulled".

Does **not** un-freeze existing per-user `DEFAULTMODEL` rows (Claude/Gemini leftovers) — that is PR C. Does **not** add local Whisper — that is PR D.

## Test plan
- [x] `make -C backend test` — 4093 tests
- [x] `make -C frontend lint` / `make -C frontend test` — 1276 tests
- [x] Browser: suggested-models button hidden when not all keys; admin catalog badges; picker uses filtered list

### DIFF
--- a/backend/src/AI/Credential/ChatReadinessService.php
+++ b/backend/src/AI/Credential/ChatReadinessService.php
@@ -6,6 +6,7 @@ namespace App\AI\Credential;
 
 use App\AI\Service\OllamaModelInventory;
 use App\AI\Service\ProviderRegistry;
+use App\Model\ModelCatalog;
 use App\Repository\ModelRepository;
 use App\Service\ModelConfigService;
 use Psr\Cache\CacheItemPoolInterface;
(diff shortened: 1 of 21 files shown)


orgaralf:
PR B of #1586. Existing per-user DEFAULTMODEL rows (Claude/Gemini leftovers) are **not** cleared here — that is the next PR. Local Whisper catalog entry is the one after that.

PR #1599, opened 2026-08-27
feat(backend): redacted prod event ring for dev/AI troubleshooting

FExB17:
## Summary

Give developers a way to see recent production errors and notable operational events **on demand**, without ever exposing raw logs (which carry chat content, user emails, RAG/document text and secrets).

The approach is deliberately **allow-list, not deny-list**: events are assembled from a fixed set of structured fields, so "nothing from the user leaks" is true *by construction*. The only free-text fields (log message, exception message) pass through a scrubber that masks emails, bearer tokens, provider keys and sensitive `key=value` pairs.

Two consumers, two paths (only the AI path is built here; SSH stays a people/trust decision):

| Consumer | Access | Sees |
| --- | --- | --- |
| Human dev (deep debugging) | SSH (out of scope) | everything — their responsibility |
| AI (in-chat help) | `recent_errors` MCP tool → admin API → redacted ring | redacted events only |

## What's included (8 commits)

1. **Request correlation id** — `X-Request-Id` on every response + stamped onto every log record via a Monolog processor. Upstream ids accepted only when short/opaque so a client can't inject PII.
2. **Redacted event ring** — `EventRingStore` (bounded, TTL'd Redis sorted set) + `EventScrubber`. Adds `zCard`/`zRemRangeByRank`/`zRevRangeByScore` to `RedisService`.
3. **Monolog handler** — `EventRingHandler` writes `warning+` records into the ring at its own level (so operational warnings stay queryable even when stderr is errors-only), enriched with route/method/correlation id; exceptions get class + scrubbed message + compact `file:line` stack.
4. **Admin API** — `GET /api/v1/admin/logs` (`ROLE_ADMIN`): `mode=recent` (level/since/query/request_id filters) and `mode=summary` (counts by level/event/route). Fully OpenAPI-annotated.
5. **MCP tool** — admin-only `recent_errors` (`summary` + filtered `recent`). Gated in the factory (hidden from non-admins in `tools/list`) **and** re-checked in the handler, because the `mcp` firewall only requires an authenticated key.
6. **Static-analysis type tightening** for the store.
7. **Generated-schema fix** — the summary count maps used OpenAPI `additionalProperties`, which produced an invalid `z.record(z.number().int())` under Zod v4; declared as plain objects instead (endpoint is not consumed by the frontend).
8. **`host` field** — every event records `gethostname()` as an allow-listed `host` field, so "which cluster node?" is answerable at a glance. Non-PII, negligible cost.

## Design notes / scope

- **Persistence:** Redis ring (volatile, ~2000 entries, 7-day TTL) — no schema migration, no new service. A `BERRORLOG` table remains a later option if weeks-long forensic retention is ever needed.
- **`user_id`** is the only quasi-identifier kept (pseudonymous, resolvable only via the access-controlled DB); email/name are never stored.
- **Coverage:** catches 500s, worker crashes, fallbacks, rate/latency anomalies, health degradation (~80–90% of incidents). Content-specific bugs ("why did *this* message fail") are intentionally out of scope — that's the human SSH case. Richness grows with instrumentation.
- **Scrubber is best-effort by design** — the allow-list is the real guarantee; the scrubber only guards the two free-text fields, and both consumers are `ROLE_ADMIN`. If zero residual is ever required, the clean step is to drop `exception_message` from the AI path (keep class + stack), not more regex.
- **GDPR:** the design applies the standard techniques (minimisation, pseudonymisation, scrubbing, RBAC, retention limits). "Compliant" still requires an org-level legal basis + DPA with the AI provider — not something code can assert.

## Follow-ups (intentionally not in this PR)

- Instrument info-level operational events at interesting spots (fallbacks, model switches) via a thin recorder — this is what turns the tool from an error list into an operational overview (request/fallback counts).
- Merge recent `BUSELOG` AI-failures (`BSTATUS='error'`) into the feed.
- Optional Messenger DLQ reader.

## Test plan

- [x] `make -C backend phpstan` (src + tests) — no errors
- [x] `make -C backend test` — full suite green (4173 tests, 0 failures/errors)
- [x] New tests: `RequestIdGeneratorTest`, `RequestIdHeaderTest`, `EventScrubberTest`, `EventRingStoreTest` (incl. host round-trip), `EventRingHandlerTest` (incl. host), `AdminLogsControllerTest`, `McpControllerTest` (admin-visibility)
- [x] OpenAPI spec builds; frontend generated schemas unchanged (backend-only change)
- [ ] Manual: hit `GET /api/v1/admin/logs?mode=summary` as admin; call `recent_errors` via MCP with an admin key


cristiangrxs:
## Review pass — findings and fixes (pushed)

Merged `main` in (the `McpServerFactory` conflict was resolved by hand, keeping both the desktop-agent tools and `recent_errors`). The overall design holds up: allow-list over deny-list is the right call, the admin gate is doubled, the server is rebuilt per request so there is no cross-user tool leak, and impersonation does not carry `ROLE_ADMIN`. Five things needed fixing.

### 1. Redaction bypass in the scrubber (security)

`scrub()` applied the length cap *after* the patterns. When `preg_replace` fails — PCRE gives up on pathological input and returns `null` — the old code silently kept the untouched text, so an email in a long exception message reached the AI-facing feed verbatim.

Now it truncates first (bounded matching cost) and fails closed to `[redacted]` when a pattern cannot be evaluated. `testFailsClosedWhenAPatternCannotBeEvaluated` is a real regression guard — against the previous implementation it fails with `+'User victim@example.com not found'`.

### 2. `since_minutes` produced a 500

The window was unbounded, so `time() - $minutes * 60` overflowed to a float and hit `TypeError: EventRingStore::recent(): Argument #2 ($sinceTs) must be of type ?int, float given` under `strict_types`. Now clamped to 7 days — the ring's TTL, and already the cap the MCP tool used.

### 3. Every allow-listed context field was dead

The OpenAPI schema and the tool description promise `user_id`, `provider`, `model`, `status_code`, `worker`, `duration_ms`, but the handler discarded the whole Monolog context, so all of them were always `null`. Those keys are already populated across the codebase (`error` 497x, `user_id` 149x, `provider` 52x, `model` 41x), so an event read as "AI request failed" with no reason at all.

They are now read by name — an allow-list, never a passthrough, since contexts also carry `email`/`to` and free-form payloads — plus `context['error']` as the scrubbed reason when no exception object is attached, which is the most common logging shape here.

### 4. A Redis outage cost up to 2.5s per warning

The reentrancy guard stopped the recursion but not the latency: Predis reconnects per command with a 2.5s timeout, on the error path. `record()` now reports whether it reached Redis and the handler mutes itself for 60s after a refusal.

### 5. Wasted work per record

`AbstractProcessingHandler::handle()` renders `$record->formatted` through a `LineFormatter` that `write()` never reads. Switched to `AbstractHandler`. The read path also no longer re-runs the scrubber over text that was already masked on write (that was up to 2000 re-scrubs per filtered query).

### Smaller

- Dropped `public: true` on `EventRingStore` — `framework.test` already exposes private services to the test container, and no other service in the repo needs it.
- The MCP gate used `User::isAdmin()` (internal user level only) while the HTTP endpoint uses `ROLE_ADMIN`, which is also granted through the OIDC role mapping. An OIDC admin could call the API but never saw the tool. Both use `ROLE_ADMIN` now.
- `EventScrubber` / `RequestIdGenerator` are now `final readonly`.
- Added `docs/OBSERVABILITY.md` (linked from the README and `docs/ADMIN.md`).

### Deliberately not changed

- The upstream `X-Request-Id` is accepted from any client, not just a trusted proxy. The charset and length limits make header injection impossible and the value stays opaque; `trusted_proxies` is not configured, so gating on it would disable propagation entirely. Documented instead, including that a correlation id is therefore not evidence of anything.
- Whether the ring is shared across `web1`/`web2`/`web3` depends on the platform's Redis topology, which is not visible from this repo. With a per-node Redis you only see the node that served your request — documented as a caveat next to the `host` field.

### Gate

`make -C backend lint`, `make -C backend phpstan` (src + tests) and the full PHPUnit suite (4571 tests, 0 failures) are green, plus `generate-schemas` + `vue-tsc` for the changed OpenAPI annotations. Mobile impact still classifies as `backend-only`.

PR #1607, opened 2026-08-27
feat(chat): keep conversation files in context for follow-up requests

orgaralf:
## Summary

Asking to *"change the color"* of a picture produced a brand new image instead of editing it. A generated file only hangs off `BFILES` via `BMESSAGEID` and never off the message relation, so every consumer that looked at the current message alone saw no file at all and treated the follow-up as a fresh generation. Documents had the same gap in a milder form.

- **`ConversationFileCatalog` is now the single resolver** for "which files does this conversation have": message attachments, `BFILES` rows by message id, and legacy `BFILEPATH` — validated against the upload root, deduplicated, categorised, with per-category budgets so a burst of generated documents cannot push the picture the user means out of the catalog. `DocumentImageCatalog` delegates to it (and gains its legacy-path fallback) instead of keeping a second resolver.
- **Cross-turn image edits work.** `MessageSorter` annotates history turns that carried a file, `MessageClassifier` passes the sorter's `BINPUTMODE` vote through and stops fast-pathing short follow-ups after a media turn, and `MediaGenerationHandler` resolves the newest conversation image as the edit source and switches to the `PIC2PIC` model. The `mediamaker` prompt is told it is editing, so it writes only the change instead of re-describing the scene. The UI shows an "Editing car-sunset.png…" status while this runs.
- **Generated images can reach a vision chat model**, behind `FILE_CONTEXT.VISION_INCLUDE_GENERATED` (seeded `0`) and only when the selected model already reads images. Off by default because each image rides along as a base64 payload on every following request of the conversation.

Implements `_devextras/planning/20260827-chat-session-file-context/`. Additive only — no schema change; the existing tables already hold everything the catalog needs.

Mobile impact: **ota-candidate** (`frontend/**` only; the backend half is `backend-only`).

## Deferred to a follow-up

Recorded in the epic `STATUS.md`: the persisted "edited from &lt;file&gt;" badge on the answer bubble (needs message-meta persistence plus an API/Zod surface), quoting an image to pin it as an explicit composer reference, and the nightly-tier Playwright flow.

## Test plan

- [x] `make -C backend phpstan` — no errors (full project, `src/` + `tests/`)
- [x] `make -C backend test` — 4272 tests, unfiltered; routing characterization snapshots did **not** drift
- [x] `make -C frontend lint`
- [x] `make -C frontend test` — 1296 tests
- [x] `node scripts/mobile-impact.mjs --base origin/main --head HEAD` → `ota-candidate`
- [ ] Manual: generate an image, then ask to recolor it — the second picture is an edit of the first, not a new composition
- [ ] Manual: generate a `.docx`, then ask for an edit — prior content is still re-injected
- [ ] Manual: flip `FILE_CONTEXT.VISION_INCLUDE_GENERATED=1` on a vision model and ask a question about a generated picture


orgaralf:
Both review comments were valid and are addressed in 403c029.

**Skewed prompt on a media-type mismatch** — the real one. The edit source is resolved before extraction so the mediamaker can be told it is editing a picture, and dropping the source afterwards left the "describe only the change" wording in a prompt that was then used for video or audio. It is now re-extracted without the edit context in that case, and `/vid` / `/tts` veto the edit source up front (the classification carries no `media_type` for a slash command, so the topic has to say it). Covered by two new tests.

**Detached docblock** — moved back onto `buildMultimodalContent()`.

Gate re-run: `make -C backend lint`, `make -C backend phpstan`, `make -C backend test` (4274 tests, unfiltered) all clean.

PR #1617, opened 2026-08-28
ci: build the arm64 release image on a native runner

cristiangrxs:
## Summary

The arm64 release image was built with `linux/arm64` under QEMU on an x86 runner. Measured on recent `main` pushes, that step took **157-173s** and was the single longest step of the release path. GitHub's arm64 runners (`ubuntu-24.04-arm`) are free and unlimited for public repositories, so the emulation buys nothing.

Simply swapping the runner is not enough: keeping one job would serialize the amd64 push behind the arm64 build and save almost nothing. The publish path becomes three jobs instead, with the two builds in parallel:

| Job | Runner | Role |
|---|---|---|
| `docker-push-amd64` | `ubuntu-latest` | Publishes the exact tar E2E ran against |
| `docker-push-arm64` | `ubuntu-24.04-arm` | Native arm64 build (in parallel) |
| `docker-manifest` | `ubuntu-latest` | Joins both digests, then verifies the result |

**Invariants are preserved.** Both build jobs keep the full gate list and the `main`/tags condition, so arm64 is still built only once every check is green, and `docker-manifest` remains the only job that moves a release tag. Its existing verification is unchanged: it refuses any tag that does not resolve to exactly the two digests just published.

### Two correctness fixes come with the split

- **`APP_VERSION` is resolved once**, by `docker-build`, and consumed downstream. The independent resolutions could previously bake *different* versions into the amd64 and arm64 images if a tag landed between them.
- **Separate Actions cache scopes per architecture.** The two builds share no layers and were evicting each other from the 10 GB cache — a plausible contributor to `docker-build` swinging between 53s (warm) and ~3.5min (cold).

`docker-push-amd64` needs neither a checkout nor the frontend artifacts; it loads a tar and pushes it.

### Expected effect

Release path after green gates: **~6min -> ~3-4min**. No change to the PR gate, and no cost either way (both runner types are free for public repositories).

## Notes

`ubuntu-24.04-arm` works in **public repositories only**. That constraint is documented in the job comment: if this repository ever becomes private, the label stops resolving and the job would queue indefinitely.

## Verification

- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated

`actionlint` reports the same findings as `main` (1 pre-existing runner-label, 8 pre-existing shellcheck) — no new ones. The job graph, every `needs` reference and every `needs.*.outputs.*` key were validated programmatically.

## Test plan

- [x] Job graph valid; no dangling `needs` or missing job outputs
- [x] `Publish arm64 Image` is green on this PR: 53s native vs 157-173s emulated, cold cache (run 33153161329, see comment below)
- [x] Temporary validation commit removed again (it and its revert cancelled out exactly, so the rebase onto main dropped both); `docker-push-arm64` is `skipped` on the final PR run
- [x] Rebased onto main after #1616 merged; the single conflict (the concurrency comment, which both PRs edited) was resolved by keeping both sides
- [ ] After merge: first `main` push publishes a multi-arch manifest, `Verify manifest and platforms` green, publish phase noticeably shorter

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ permissions:
 # Superseded runs on the same PR just burn runner capacity, which is shared
 # across the organization (the plan allows 60 concurrent standard-runner jobs
 # and one run peaks at ~7). Push/tag runs get a unique group (run_id) so main
-# runs neither serialize nor cancel each other — docker-push and
+# runs neither serialize nor cancel each other — the publish jobs and
 # cloudflare-deploy must never be interrupted or skipped.
 concurrency:
   group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
@@ -489,6 +489,12 @@ jobs:
     outputs:
       image-tags: ${{ steps.meta.outputs.tags }}
       image-labels: ${{ steps.meta.outputs.labels }}
+      # The version baked into the image this job builds and every downstream
+      # job publishes. Resolved once, here, so the amd64 image, the arm64
+      # image and the version advertised to deployments can never disagree —
+      # independent resolutions would drift the moment a tag lands between
+      # them.
+      app-version: ${{ steps.app-version.outputs.version }}
 
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


cristiangrxs:
## Native arm64 build validated

The temporary build-only commit ran the job for real on `ubuntu-24.04-arm` ([run 33153161329](https://github.com/metadist/synaplan/actions/runs/33153161329)) and it is green, with a **cold cache** (this was the first build under the new `scope=arm64`):

| | QEMU on x86 (today) | Native `ubuntu-24.04-arm` |
|---|---|---|
| `Build and push arm64 image` step | 157-173s | **53s** (~3x faster) |
| Job total incl. runner start, checkout, artifacts | — | 77s |

Everything that could have differed from the emulated build went through natively: `composer proto:generate`, `composer install --no-dev --classmap-authoritative`, and the `bookworm-backports` libheif packages. The base image `synaplan-base-php:1.2.0` is pinned by its manifest-list digest and resolved its arm64 variant as expected.

The temporary commit is now reverted (53c4426cd). `docker-push-arm64` is back to `push: true` and to main/tags only, so it is `skipped` on this PR from here on — that is the intended final state, and the evidence above is the record.

PR #1648, opened 2026-08-31
chore(release): install 4.4.1 in new deployments

cristiangrxs:
## Summary

- Installs `4.4.1` as the default version for **new** deployments (`elestio.yml`, `deploy/selfhost.env.example`, Umbrel App Store package including the published digest, AWS Packer AMI).
- Existing self-host / Umbrel / AWS installations keep the version their operator pinned. A redeploy never moves them.
- **Hosted production (`synaplan-platform`) is not updated by this PR.** That stack has no `APP_VERSION` in `.env` (correct — the running image bakes the version). It still shows 4.4.0 because the containers have not pulled a newer image. Do **not** add `APP_VERSION` to the prod `.env`. After a backup, on each node: `docker compose up --pull always -d`. To stay on the release rather than later `main` commits in `:latest`, pin `ghcr.io/metadist/synaplan:4.4.1`.

## Elestio

Merging this restarts every Elestio instance that tracks `main`, because Elestio redeploys on each commit to the branch it tracks. Merge when that is convenient.

The Umbrel pin lives in this repository only. Submitting the raised package to [getumbrel/umbrel-apps](https://github.com/getumbrel/umbrel-apps) remains a separate, manual pull request.

## Test plan

- [ ] Diff is only the five pin files, all at `4.4.1` (Umbrel image `4.4.1@sha256:38c2ba8fde0ae685060299567c02413876ca5dba391e6fa94fa80b6131ce0615`)
- [ ] Confirm hosted prod is updated by image pull, not by merging this PR or editing `.env`

### DIFF
--- a/deploy/aws/packer/synaplan.pkr.hcl
+++ b/deploy/aws/packer/synaplan.pkr.hcl
@@ -23,7 +23,7 @@ variable "synaplan_version" {
   # Raised by scripts/set-release-version.mjs after every published release, so
   # a build with no arguments produces the AMI for the release this branch
   # ships. The release workflow still passes it explicitly.
-  default     = "4.4.0"
+  default     = "4.4.1"
   description = "Released SemVer version to bake in, without a leading v. Never a mutable tag: the first boot pins deploy/.env to exactly this value."
 
   validation {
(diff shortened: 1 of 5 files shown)


cristiangrxs:
Duplikat von #1655, das auf `automation/default-release-version` liegt — dem Branch, den `release-rollout.yml` beobachtet. Der Pin auf 4.4.1 rollt über #1655 aus.

PR #1649, opened 2026-08-31
fix(ci): open the release-version pin proposal reliably

cristiangrxs:
## Problem

After `v4.4.1` the `release-version-pin` job pushed the raised pins to `automation/default-release-version` and then failed to open the pull request:

```
GraphQL: Resource not accessible by integration (createPullRequest)
```

It reported this as `::warning::` on a **green** job, so nobody noticed. `main` kept pinning 4.4.0 for new deployments for two days. The pin PRs have in fact always been opened by hand ([#1622](https://github.com/metadist/synaplan/pull/1622) was authored by a human, not the automation).

Root cause: `gh pr create` ran with the release App token, and that App is scoped to Contents only.

## Fix

`gh pr create` now tries two tokens and fails loudly if neither works:

1. **Release App token** — preferred, because CI only triggers on `pull_request`, and a PR created with `GITHUB_TOKEN` starts no workflow run at all.
2. **`GITHUB_TOKEN`** — fallback (the job already has `pull-requests: write`). The PR opens but carries no checks, so the step emits a `::warning::` and the step summary says to close/reopen it before merging.
3. **Neither** — `::error::` and `exit 1` instead of a green job with a buried warning.

Tokens are passed to the helper by variable name (`${!1}`), so no secret goes through argv.

Without checks, a pin PR would skip exactly the jobs that validate it: `Deployment Templates (AWS)` (`cfn-lint`, `packer validate`) and `Mobile Release Tooling` (`node --test tests/*.test.mjs`, which contains the "the shipped files name one and the same released version" pin-consistency test).

## Verification

The step's `run` block was extracted from `ci.yml` and executed against stubbed `gh`/`git`/`node`:

| Case | Result |
| --- | --- |
| App can create | exit 0, PR with checks, no warning |
| App blocked, `GITHUB_TOKEN` works (today) | exit 0, PR opened, warning about missing checks |
| App not configured, `GITHUB_TOKEN` works | exit 0, same warning |
| Both blocked | exit 1, `::error::` + compare URL |

`node --test tests/set-release-version.test.mjs` passes (18/18).

## Scope

The missing 4.4.1 pin itself is proposed separately in #1648. This PR only changes how the proposal gets opened.

This does **not** make any deployment auto-update: it only guarantees the proposal appears. Merging, the Umbrel store submission, and the AWS Marketplace submission stay manual.

Hosted `synaplan-platform` is unaffected — it has no `APP_VERSION` in `.env` (correct; the running image carries the version) and updates only by pulling a newer image.

## Test plan

- [ ] Review the two-token chain in the `Propose the new default version` step
- [ ] Next tagged release opens `chore(release): install X.Y.Z in new deployments` automatically
- [ ] If it opens via `GITHUB_TOKEN`, confirm the step summary tells the maintainer to close/reopen for checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1521,10 +1521,22 @@ jobs:
             "https://x-access-token:${PUSH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" \
             "HEAD:refs/heads/${BRANCH}"
 
-          existing="$(gh pr list --head "$BRANCH" --state open --json number --jq '.[0].number // ""')"
-          if [ -n "$existing" ]; then
-            echo "Updated [pull request #${existing}](${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/pull/${existing}) to \`${VERSION}\`." |
+          # Non-fatal on purpose: under `set -e` a bare assignment from a
+          # failing lookup would end the step right here — after the push, but
+          # before any of the reporting below — leaving a red job with no
+          # explanation of what to do.
+          find_open_pr() {
+            gh pr list --head "$BRANCH" --state open --json number --jq '.[0].number // ""' || true
+          }
+
+          report_existing_pr() {
+            echo "Updated [pull request #${1}](${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/pull/${1}) to \`${VERSION}\`." |
               tee -a "$GITHUB_STEP_SUMMARY"
+          }
+
+          existing="$(find_open_pr)"
+          if [ -n "$existing" ]; then
+            report_existing_pr "$existing"
             exit 0
           fi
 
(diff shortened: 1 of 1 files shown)


cristiangrxs:
Addressed the suppressed review comment about `gh pr list` under `set -euo pipefail` in 01488cb5b — it was a real hole, and fixing it naively would have opened a second one.

**The reported problem:** a bare assignment from a failing lookup ends the step immediately, *after* the force-push but *before* any of the reporting. The maintainer would get a red job with a raw `gh` error and no compare URL.

**Why `|| true` alone is not enough:** with an empty `existing`, the step proceeds to create a PR. If one already exists (which is exactly the state a failed lookup hides), both creates fail with `already exists` and the step would report a red release even though everything worked — the force-push had already carried the new version into that proposal.

**What the fix does:**

- `find_open_pr()` is non-fatal, so the lookup can never end the step early.
- After both creates fail, the lookup is retried once; if a proposal is found it is reported as carried forward.
- If the lookup is *persistently* unavailable, the `already exists` error from `gh pr create` is treated as the intended outcome, and the existing proposal's URL is extracted from that error for the step summary.
- Only a genuine inability to open or find a proposal is `::error::` + `exit 1`.

I did not add a token retry for the lookup itself: `GITHUB_TOKEN` has `pull-requests: write` on this job, so a failure there is transient or a hard restriction, and both are now handled without depending on the lookup succeeding.

**Verified** by extracting this step's `run` block from `ci.yml` and executing it against stubbed `gh`/`git`/`node` (the `gh` stub mirrors the real `already exists:` + URL output), across seven cases:

| Case | Expected | Result |
| --- | --- | --- |
| App can create | exit 0, PR with checks | pass |
| App blocked, `GITHUB_TOKEN` works | exit 0 + "no checks" warning | pass |
| App not configured | exit 0 + warning | pass |
| Both tokens blocked | exit 1 + `::error::` | pass |
| Proposal already exists | exit 0, reported as updated | pass |
| Lookup fails, no PR yet | exit 0, PR created | pass |
| Lookup fails **and** PR exists | exit 0, existing PR reported | pass |

PR #1652, opened 2026-08-31
feat: distribute a release to Elestio, Umbrel and AWS automatically

cristiangrxs:
## Summary

A release already produced everything needed to update the catalogs — a verified image, a manifest entry, a pull request raising all four version pins, and two AMIs — and then stopped, waiting for someone to merge and to fill in two web forms. Until that happened, anyone deploying Synaplan installed the previous release. That is how `main` kept pinning 4.4.0 after 4.4.1 shipped.

This closes the last mile in three places.

**`release-rollout.yml`** merges the pin proposal in a nightly maintenance window, not on the spot: merging restarts every Elestio instance tracking the default branch, and auto-merge would do that mid-afternoon. Five guards must agree first — the catalog pins one released version across all five anchors, that version is what the published manifest advertises, it is not yanked, the pull request touches nothing but the four pin paths, and `All Checks Passed` is green. Anything else holds the rollout back with a warning instead of merging.

**`umbrel-store-sync.yml`** carries the raised package into `getumbrel/umbrel-apps`, which cannot pull from here. It rebuilds our fork from upstream `master`, mirrors the package, runs Umbrel's own `lint:apps --check-images`, and pushes. Which pull request it lands in is decided by the store, because Umbrel's linter applies opposite rules to each case: a new submission must leave `releaseNotes` and `gallery` empty, an update to a released app is expected to carry release notes. This corrects a wrong assumption in the plan — those two fields were treated as placeholders waiting to be filled, and for a new submission they are the required values.

**`aws-ami.yml`** gains a `submit` job that offers the built and verified AMI to the listing through the Catalog API. Each architecture becomes its own version, sequentially: AWS requires all delivery options of a version to share one `AmiSource`, and a running change set locks the entity. The default intent is `VALIDATE` — a submitted version sits in AWS review for days, so a real submission is a deliberate act (`AWS_MARKETPLACE_SUBMIT=apply`), not the side effect of setting a secret.

Running installations are deliberately untouched, everywhere. They keep the version their operator pinned, because an update runs database migrations and the backup that makes those survivable is the operator's to take. The admin area's promise — it reports a newer version and never installs one — still holds.

## What still needs a human

- Merging the Umbrel store pull request (Umbrel's repository, Umbrel's review).
- AWS's review of a submitted version, 5–14 working days.
- Existing Elestio pipelines: their environment is materialized once at creation and ignores `elestio.yml` afterwards, so nothing in this repository can reach them.

## Prerequisites

The release App needs `Pull requests: write` and a bypass entry in the `ReviewCodeSyn` org ruleset, or the rollout can never merge — without checks on its proposal, guard 5 is unsatisfiable by design. New secrets: `UMBREL_FORK_TOKEN`, `AWS_MARKETPLACE_PRODUCT_ID`, `AWS_MARKETPLACE_INGESTION_ROLE_ARN`, and optionally `DISCORD_RELEASE_WEBHOOK`. New variable: `AWS_MARKETPLACE_SUBMIT`. The `synaplan-marketplace-roles` stack needs redeploying for the new catalog permissions. Every workflow skips itself with a notice while its own settings are missing.

## Test plan

- [x] `node --test "tests/*.test.mjs"` — 58 pass, including new suites for the pin reader and the release-notes writer
- [x] Umbrel's real linter against the generated package, both paths: new submission and version bump — `No issues found`
- [x] `cfn-lint deploy/aws/cloudformation/*.yaml` — clean; `aws cloudformation validate-template` accepts the template
- [x] `bash deploy/scripts/tests/test-lifecycle.sh` and `bash deploy/aws/scripts/tests/test-firstboot.sh`
- [x] Every guard's jq expression exercised against the shapes it has to survive, including a yanked entry as both a string and an object
- [x] All 97 workflow `run` blocks parse under `bash -n`
- [ ] `release-rollout.yml` via `workflow_dispatch` with `dry_run` against the next real proposal
- [ ] `aws-ami.yml` on the next release tag, to see one `VALIDATE` change set before `AWS_MARKETPLACE_SUBMIT` is switched

Independent of #1649; both touch `ci.yml` but not the same lines.

### DIFF
--- a/.github/workflows/aws-ami.yml
+++ b/.github/workflows/aws-ami.yml
@@ -214,9 +214,9 @@ jobs:
             echo ""
             echo "AMI: \`${ami}\` in \`${AWS_REGION}\`"
             echo ""
-            echo "Next: run **Test Add Version** in the AWS Marketplace Management Portal"
-            echo "for the free compliance scan, then verify a launch from"
-            echo "\`deploy/aws/cloudformation/synaplan-new-vpc.yaml\` before submitting."
+            echo "The submit job below offers this AMI to the listing once the"
+            echo "verification launch has passed. It validates by default and only"
+            echo "submits for real when the \`AWS_MARKETPLACE_SUBMIT\` variable says so."
           } >> "$GITHUB_STEP_SUMMARY"
 
       - name: Validate and share the AMI with AWS Marketplace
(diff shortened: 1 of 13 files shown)


orgaralf (approved):
Cool distribution branch!

PR #1655, opened 2026-08-31
chore(release): install 4.4.1 in new deployments

cristiangrxs:
Installs `4.4.1` in **new** deployments:

- `elestio.yml` for the one-click path
- `deploy/selfhost.env.example` for self-hosting
- `deploy/umbrel/synaplan/` for the Umbrel App Store package (version,
  `APP_VERSION`, and the published image digest
  `sha256:38c2ba8fde0ae685060299567c02413876ca5dba391e6fa94fa80b6131ce0615`)
- `deploy/aws/packer/synaplan.pkr.hcl` for the AWS Marketplace AMI

Existing deployments are unaffected. They keep the version their operator
pinned and change it only by following the update guide, so a redeploy never
moves a running installation to a different version.

Merging this restarts every Elestio instance that tracks `main`, because Elestio
redeploys on each commit to the branch it tracks. That is why nothing merges this
on the spot: `release-rollout.yml` does it in the next nightly maintenance
window, once the checks here are green and this version is still the one the
release manifest advertises. Merge it by hand to roll out sooner.

Once merged, `umbrel-store-sync.yml` carries the raised package into
[getumbrel/umbrel-apps](https://github.com/getumbrel/umbrel-apps).

Opened by hand to catch up: the automatic proposal for `v4.4.1` failed to open on
2026-08-29, because the release App could not create pull requests at the time.
That is fixed in #1649, so the next release proposes this by itself. The digest
above was read back from the published manifest list in ghcr.io, which carries
`linux/amd64` and `linux/arm64`.

### DIFF
--- a/deploy/aws/packer/synaplan.pkr.hcl
+++ b/deploy/aws/packer/synaplan.pkr.hcl
@@ -23,7 +23,7 @@ variable "synaplan_version" {
   # Raised by scripts/set-release-version.mjs after every published release, so
   # a build with no arguments produces the AMI for the release this branch
   # ships. The release workflow still passes it explicitly.
-  default     = "4.4.0"
+  default     = "4.4.1"
   description = "Released SemVer version to bake in, without a leading v. Never a mutable tag: the first boot pins deploy/.env to exactly this value."
 
   validation {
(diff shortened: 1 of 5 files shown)


orgaralf (approved):
Yes!

PR #1731, opened 2026-09-07
feat(agents): add owner CRUD and pinned agentId runtime (S1)

orgaralf:
## Summary
Assistants exist as BAGENTS rows with a validated agent.v1 draft. The owner can create them through /api/v1/agents, and a chat that carries agentId pins to that assistant without changing the unpinned classifier path.


orgaralf:
Was already merged into #1734

PR #1735, opened 2026-09-07
fix(auth): stop store redirects from cancelling the logout navigation

FExB17:
## Problem

Logout hands the browser a destination. For OIDC that is a full-page navigation to the provider's end-session endpoint, which lands on `/logged-out`. Until that navigation commits, the SPA is still alive with no user in memory while requests started before the click are still resolving.

Three call sites reacted to that momentarily missing user by assigning `window.location.href` themselves:

- `frontend/src/stores/chats.ts:23` `checkAuthOrRedirect()`
- `frontend/src/stores/history.ts:22` `checkAuthOrRedirect()`
- `frontend/src/services/api/chatApi.ts:235` `getSseToken()` catch branch

That assignment replaces the pending logout navigation instead of following it. The end-session request dies as `net::ERR_ABORTED`, the browser lands on `/login`, and the provider session stays open — the user looks logged out but is not. A chat could also still be created after the logout click, because `createChat` passed the same guard while the logout POST was in flight.

## Evidence

Playwright traces of 10 flaky runs: 9 show the chats-store redirect, 1 the SSE-token redirect. The CI list reporter prints the hijack URL in only 6 of 13 cases, which is why `?reason=session_expired` never appears in job logs even though one trace has it.

Job log of the earliest occurrence, run 32979378462, job 98213498814:

```
✘  13 [chromium] › tests/e2e/tests/oidc-auth.spec.ts:16:3 › @auth should logout from OIDC session (19.3s)
✓  14 [chromium] › tests/e2e/tests/oidc-auth.spec.ts:16:3 › @auth should logout from OIDC session (retry #1) (4.0s)
Error: expect(locator).toBeVisible() failed
Locator: locator('[data-testid="page-logged-out"]')
 - navigated to "http://localhost:8001/login?reason=auth_required"
```

Reproduced locally against a CI-like stack: with the provider end-session navigation held open by route interception, a spec forcing this race failed 3/3 without this change and passed 3/3 with it, alongside the rest of the OIDC suite.

Affected specs: `oidc-auth.spec.ts` `@auth should logout from OIDC session`, and `auth.spec.ts`, where the same aborted navigation surfaces as `page.goto: net::ERR_ABORTED at /profile`.

## Timing history

The racing navigations date back to 2026-02-21, when `window.location.href = oidcLogoutUrl` was introduced. The observed failure rate stepped up on 2026-08-26: 0 flakes in 357 consecutive OIDC jobs from 2026-08-08 to 2026-08-26 14:05Z, then 13 of 468 jobs (2.8%, rising to 3.8% in September). No commit in that window touches this path — the push carrying the first flake, `d32416207` (#1573), changes only `phpstan` and `php-cs-fixer` in `composer.lock` and is frontend-identical to a clean job 13 minutes earlier. Job logs older than roughly 30 days are expired, so June and July cannot be checked.

What shifted the timing is undetermined. This change removes the competing navigation rather than widening a wait, so it does not depend on the answer.

## Change

`frontend/src/services/sessionTeardown.ts` marks the window between the start of a logout and the browser leaving the page. Inside it, an unauthenticated state is expected rather than an error: no navigation, no protected requests.

The window opens in the auth store's `logout()` before any state is cleared, and closes when a principal is signed in again. `authService`'s own silent logouts, triggered by a refresh that came back 401, stay outside it deliberately: they never request an end-session URL, so there is no navigation to protect and their ordinary redirect to `/login` has to keep working.

## Tests

`frontend/tests/unit/stores/logoutNavigationRace.spec.ts` asserts both directions: the `auth_required` and `session_expired` redirects still fire outside a teardown, and inside one the stores neither navigate nor call the API, including `createChat` returning `null`. `auth-impersonation.spec.ts` covers the flag's lifecycle across a logout.

Frontend gate green: eslint, `vue-tsc`, 1654 unit tests.

## Mobile impact

`store-required`, from `frontend/src/stores/auth.ts` alone, which `.github/mobile-impact-policy.json:45` lists explicitly. The change there is the two teardown calls; no authentication transport or native seam is touched. The new `frontend/src/services/sessionTeardown.ts` classifies as `ota-candidate` under the existing `services/` rule, so no policy change is needed.

### DIFF
--- a/frontend/src/services/api/chatApi.ts
+++ b/frontend/src/services/api/chatApi.ts
@@ -8,6 +8,7 @@ import { isNativeApp } from './nativeRuntime'
 import { getNativeAccessToken, hasNativeTokens } from './nativeAuth'
 import { UserMemorySchema } from './userMemoriesApi'
 import { hasSessionHint, clearSessionHint } from '@/services/sessionHint'
+import { isSessionTerminating } from '@/services/sessionTeardown'
 import { GetApiChatsMessagesResponseSchema } from '@/generated/api-schemas'
 import type { StreamUpdatePayload } from '@/types/chatStream'
 
@@ -232,8 +233,14 @@ async function getSseToken(): Promise<string | null> {
 
       return cachedSseToken
     } catch (error) {
-      // If error is "Authentication required", redirect to login
-      if (error instanceof Error && error.message === 'Authentication required') {
+      // If error is "Authentication required", redirect to login — unless a
+      // logout is already under way, whose navigation this would cancel
+      // (see sessionTeardown).
+      if (
+        error instanceof Error &&
+        error.message === 'Authentication required' &&
+        !isSessionTerminating()
+      ) {
         // Trigger auth failure handling (redirect to login)
         window.location.href = `/login?reason=session_expired`
       }
--- a/frontend/src/services/sessionTeardown.ts
+++ b/frontend/src/services/sessionTeardown.ts
@@ -1,0 +1,43 @@
+/**
+ * Session Teardown - marks the window in which the client is giving up its
+ * session but the SPA is still alive.
+ *
+ * Logout is not instant: it awaits POST /api/v1/auth/logout, wipes the
+ * user-scoped stores, disconnects realtime, and only then hands the browser
+ * over to the next page - for OIDC a full-page navigation to the provider's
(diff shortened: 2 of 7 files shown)


orgaralf (approved):
Go ahead, contains non-blockers for my PRs... thanks!

PR #1785, opened 2026-09-10
fix(frontend): keep the composer focused after sending a message

FExB17:
## Problem

After sending a message you had to click back into the input field before
you could type the next one — focus was lost on every send.

Two causes, depending on the surface:

1. Submitting a form (or clicking the send button) moves focus **onto that
   button**. The button is then disabled while the request runs, so focus
   falls back to `<body>`.
2. Several composers additionally set `:disabled` on the input for the
   duration of the request, and the browser blurs a disabled element.

Reported for the widget chats inbox (`/channels/widgets/:widgetId/chats`),
but the same bug existed on every send surface.

## Fix

Refocus the input after the send is dispatched.

| Surface | File | When |
| ------- | ---- | ---- |
| Widget chats inbox | `views/WidgetSessionsView.vue` | `finally` + `nextTick` (input is disabled while sending) |
| Live support | `views/LiveSupportView.vue` | `finally` + `nextTick` |
| Widget AI setup panel | `components/widgets/WidgetAiSetupPanel.vue` | `finally` + `nextTick` |
| Main chat composer | `components/ChatInput.vue` | synchronous |
| Embeddable chat widget | `components/widgets/ChatWidget.vue` | synchronous |
| Assistant test panel | `components/assistants/AssistantTestPanel.vue` | synchronous |

Where the field stays enabled during the request, the refocus happens
synchronously inside the click gesture so the **mobile keyboard stays open**
as well. Where the field is disabled, it has to wait for the re-enable —
that is the pattern `components/widgets/SetupChatModal.vue` already used,
which was the only composer that got this right.

## Tests

- New `tests/unit/components/ChatInputFocus.spec.ts` — focus is retained
  after sending via the button and via Enter.
- `tests/unit/components/assistants/AssistantTestPanel.spec.ts` — added a
  focus assertion.

All three new tests were verified to fail without the production change.

## Verification

- `make -C frontend lint` — green (only pre-existing warnings in an
  untouched file)
- `docker compose exec -T frontend npm run check:types` — green
- `make -C frontend test` — 1751 tests, all green
- `make test-e2e` — 17 pre-existing failures in this sandbox (no AI
  provider key: email, subscription, WhatsApp, guest chat, widget). A
  baseline run of the same specs on unmodified `main` produced the same
  failures (18 — one more, flaky), so this change introduces no E2E
  regression.

Mobile impact: `ota-candidate` (all changed source paths are under
`frontend/src/**`; no policy allow-list change needed).

### DIFF
--- a/frontend/src/components/ChatInput.vue
+++ b/frontend/src/components/ChatInput.vue
@@ -1013,6 +1013,11 @@ const sendMessage = () => {
   clearPersistedInput()
   clearPersistedAttachments()
   clearPersistedBlocks()
+
+  // Sending via the button moves focus onto that button, so the next message
+  // would need a click back into the composer. Refocus synchronously — inside
+  // the click gesture — so the mobile keyboard stays open as well.
+  textareaRef.value?.focus()
 }
 
 const toggleThinking = () => {
--- a/frontend/src/components/assistants/AssistantTestPanel.vue
+++ b/frontend/src/components/assistants/AssistantTestPanel.vue
@@ -19,6 +19,7 @@
     </div>
     <form class="flex gap-2" @submit.prevent="send">
       <input
+        ref="draftInputRef"
         v-model="draft"
         type="text"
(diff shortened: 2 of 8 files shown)


orgaralf (approved):
OK

PR #1801, opened 2026-09-10
fix(search): log and surface which web search provider answered

### LINKED ISSUE #1791: Web search: the provider that answered is never observable, and a failing provider is indistinguishable from an empty result
## Problem

Nothing tells an operator which web search provider actually answered. The information exists in the result set — including the fact that a fallback took over — but it is dropped before it reaches any surface an operator can look at.

---

## Expected

The provider that served a search, and a fallback if one occurred, are observable without reading source code: in the admin test result, in the search metadata returned by the API, and in a log line for the normal (non-fallback) case.

This is about the operator's diagnostic path, not the end user's result list — a chat user has no use for the provider name and should not need to see it.

## Actual

`WebSearchRegistry::search()` records both facts in the result meta:

```106:112:backend/src/Plug/WebSearch/WebSearchRegistry.php
        $fallback = $this->byKey[$fallbackKey] ?? null;
        $fallbackSet = $this->trySearch($fallback, $query, $fallbackKey);
        if (null !== $fallbackSet) {
            $this->fallbackMetrics->increment($activeKey, $fallbackKey);

            return $fallbackSet->withMeta(['fellBackFrom' => $activeKey]);
        }
```

Measured on the dev stack with an invalid Exa key as active provider and Brave as fallback:

```
meta: {"total":10,"provider":"brave","latencyMs":644,"fellBackFrom":"exa"}
count: 10
```

Neither field is forwarded. The SSE event `search_complete` carries query and result cards only (`MessageProcessor` L481–488), and the message metadata carries query and result count only (`MessageApiFormatter` L112–119). The admin **Test query** panel returns titles, latency and error, but not the provider that produced them.

Three states are therefore indistinguishable from the outside:

| State | What an operator sees |
| ----- | --------------------- |
| Configured provider answered | results |
| Configured provider failed, fallback answered | results |
| Configured provider failed, no fallback configured | no results, **no error** — the empty set is intentional (`WebSearchRegistry::search()`, "Failures become an empty set — callers never see the exception (C7)") |

The third state was reproduced with an invalid key and no fallback: `count: 0`, `meta: {"provider":"exa"}`, no exception, no log line naming a failure. "The web has nothing on this" and "the provider is broken" look identical.

---

## Steps to reproduce

1. Store an invalid API key for a provider and select it as the active provider (`/admin/setup` → **Web search**).
2. Leave the fallback empty, run a web search from chat — the answer is produced without any web results and without an error.
3. Set the fallback to a working provider, repeat — results appear, with nothing indicating that a different provider produced them.

---

## Notes

A fallback *is* counted (`WebSearchFallbackMetrics::increment()` logs `synaplan_plugs_web_search_fallback_total`), so the fallback case is at least visible in the backend log. The successful primary case and the silent-empty case have no such marker.

Related: the health line on each provider card reports configuration rather than reachability, so the card cannot be used to identify the failing provider either (filed separately).

## Direction

Forward `provider` and `fellBackFrom` from the result meta into the admin test response, and log the resolved provider for every search, not only for fallbacks. Exposing it in the user-facing chat result is not required.

orgaralf:
## Summary
- Admin Test query returns `provider` and `fellBackFrom`. The Web search tab shows “Answered by {provider}” (and the fallback source when one is set). Chat users still do not see the provider name.
- Every registry search logs `synaplan_plugs_web_search_resolved` (provider, result count, fallback, empty). Adapter throws and unconfigured skips log `synaplan_plugs_web_search_failed`. A broken provider is no longer the same log line as “no hits”.
- Empty-result warnings in `MessageProcessor` include the same meta. Chat SSE / user-facing cards are unchanged.

Stacked on #1800 (which is stacked on #1799).

Fixes #1791

## Test plan
- [x] `make -C backend lint` / `phpstan` / `test` (5975)
- [x] `make -C frontend lint` / `vue-tsc` / `test` (1759)
- [ ] Admin → Web search → Test query → line names the provider
- [ ] Chat search with a down primary and working fallback → log `resolved` with `fellBackFrom` + existing fallback counter
- [ ] Invalid key, no fallback → log `failed` then `resolved` with `empty: true`; chat still has no provider label

Classification: **ota-candidate** (`frontend/src/**`).


orgaralf:
Copilot review follow-up:

**Accepted**
- Empty chat outcomes no longer set `fellBackFrom` when the fallback also fails. Failed logs still name both providers.
- The admin Test panel shows “Answered by …” only when `error` is null, so a failed probe does not look like a successful answer.

**Declined**
- Routing admin Test through `WebSearchRegistry::search()` would hide the selected radio behind automatic fallback. Test is intentionally single-provider.

Vendor names stay off chat cards / SSE `search_complete` / user-facing `webSearch`.

PR #1804, opened 2026-09-10
fix(extraction): honour the saved adapter chain and match Test upload to chat

### LINKED ISSUE #1792: AI Plugs: extraction chain editor has no effect on built-in adapters
## Problem

The Extraction tab (`/admin/setup?tab=extraction`) offers an ordered, editable adapter chain for six file families, but only non-built-in adapters are ever read from it. The nine built-in keys are filtered out before execution and run in a hard-coded order, so reordering, removing, or even emptying a chain changes nothing — while the API reports success and the UI shows the saved order.

---

## Expected

Moving an adapter up, removing it, or emptying a family chain changes which adapter extracts the file. An adapter that is not in the chain does not run.

## Actual

`GET /api/v1/admin/plugs/extraction` offers ten keys: `native, structured_office, office_convert, tika, pdf_vision, vision, stt_cloud, whisper_local, video_analysis, docling`. Exactly one of them — `docling` — is honoured. The other nine run regardless of the stored chain.

Five counter-tests, each written through `PUT /api/v1/admin/plugs/extraction/chains`, stored value verified in `BCONFIG`, then a real chat upload (`POST /api/v1/messages/upload-file`):

| Stored chain | Log evidence | Winner |
| --- | --- | --- |
| `document = ` (**empty**, saved with HTTP 200) | `FileProcessor: Tika extraction success {"strategy":"tika","bytes":214}` | `tika` |
| `document = pdf_vision` (Tika removed) | same as above | `tika` |
| `document = tika,structured_office,office_convert,pdf_vision,docling` | `FileProcessor: extra extractor failed, falling through {"key":"docling"}` logged **before** the Tika call | `docling` still attempted first |
| `audio = whisper_local,stt_cloud` | `FileProcessor: Trying external STT provider first` → `strategy: whisper_api` | cloud STT |
| `image = ` (**empty**) | `PreProcessor: Image processed with Vision AI` | vision still ran |

The empty-chain rows are the clearest case: a family with no adapters at all is accepted, stored as an empty string, and the file is still extracted.

## Steps to reproduce

1. `/admin/setup` → tab **Extraction** → block *Documents*: remove every entry, then **Save chain**. The toast reports success.
2. Confirm the stored value: `SELECT BVALUE FROM BCONFIG WHERE BOWNERID=0 AND BSETTING='EXTRACTION.CHAIN.document';` → empty.
3. Upload any text-layer PDF in a chat.
4. `docker compose logs --since 60s backend | grep -a 'Chat file extracted'` → `"strategy":"tika"`.

---

## Root cause

`PlugConfigService::BUILTIN_EXTRACTOR_KEYS` (`backend/src/Plug/PlugConfigService.php:82`) lists `native, structured_office, office_convert, tika, pdf_vision, vision, stt_cloud, whisper_local, video_analysis`. `extraExtractorKeys()` (`:133`) strips exactly those keys from the configured chain, so `FileProcessor::tryExtraExtractors()` (`backend/src/Service/File/FileProcessor.php:202`) only ever sees the remainder. Everything after that runs in the fixed sequence hard-coded in `FileProcessor::extractText()` (`:152`) and `extractAfterOfficePrep()` (`:363`). For audio, the branch is chosen by `AiFacade::hasConfiguredSttProvider()` inside `extractFromAudio()` (`:1000`), not by the chain.

An empty chain survives because `readGlobal()` (`PlugConfigService.php:453`) returns the stored empty string rather than the default, and `splitList('')` (`:483`) yields an empty list — so nothing is iterated and the hard-coded cascade runs unchanged.

The comment at `FileProcessor.php:198` ("Empty in S1 (seeded chains only list built-in keys). S2 Docling lands here") shows this was the S1 design; the admin UI shipped the full editor on top of it.

## Secondary defect: images never reach FileProcessor

Chat image uploads bypass the extraction layer entirely. `MessagePreProcessor::processImageWithVision()` (`backend/src/Service/Message/MessagePreProcessor.php:606`, called from `:307` and `:517`) calls `AiFacade::analyzeImage()` directly — there is no `FileProcessor: Starting extraction` log entry, no chain lookup and no quality gate. This is why the `image` family survives an empty chain, and it is a separate cause from the filtering above: fixing `extraExtractorKeys()` alone will not make the `image` chain effective.

## Additional: empty chain accepted

`PUT /api/v1/admin/plugs/extraction/chains` with an empty array for a family returns HTTP 200 and stores an empty string. A family with no adapters is not a valid configuration and should be rejected.

---

## Notes

Direction: either make the chain authoritative for all keys (`FileProcessor` iterates the configured order and dispatches built-ins through the same interface the extra adapters use), or make the UI honest by rendering the built-in sequence as fixed and restricting the editor to the keys that are actually resolved. A chain editor that silently ignores nine of ten entries is worse than no editor, because an operator who removes an adapter believes it is off.

Related: #1786 (Docling multipart defect, same tab), #1793 (the Test upload panel does not reflect a real upload either).

## Verification

- With `document` set to a single adapter, an upload of a text-layer PDF logs that adapter — not `tika`.
- An empty chain is rejected on save.
- With `image = vision`, an image upload produces a `FileProcessor` log entry.
- With `audio = whisper_local,stt_cloud` and a configured cloud STT model, an audio upload logs `strategy: whisper_local`.

### LINKED ISSUE #1793: AI Plugs: the extraction Test upload panel uses a different adapter and a different AI provider than a real upload
## Problem

The **Test upload** panel in the Extraction tab is the only surface that tells an operator which adapter extracted a file. For audio and images it names a different adapter — and calls a different AI provider — than a real upload does, and it never shows that a fallback happened.

---

## Expected

The panel reflects what a real upload does, and lists every adapter that was tried with its verdict — as the service's own class comment promises: "Walks extra extractors (then FileProcessor) so the admin Test button can show every attempt: *Docling unavailable — Tika used instead*".

## Actual

The same five files, once through `POST /api/v1/admin/plugs/extraction/test` and once through a chat upload:

| File | Panel | Real chat upload |
| --- | --- | --- |
| text-layer PDF | `tika` | `tika` |
| image-only PDF | `rasterize_vision` | `rasterize_vision` |
| DOCX | `tika` | `tika` |
| MP3 | **`whisper_local`** | **`whisper_api`** (cloud provider) |
| PNG | `vision_ai` via **OpenAI `gpt-4o`** | `vision_ai` via **Groq `qwen/qwen3.6-27b`** |

The image row is the sharpest case. Both the account row and the global row of `BCONFIG.DEFAULTMODEL.PIC2TEXT` point at the same Groq vision model, yet the panel calls OpenAI:

```
# panel
AI vision request - provider selection {"requested_provider":"openai","user_id":null,"options":[]}
OpenAI: Analyzing image {"model":"gpt-4o", ...}

# real chat upload, same image
AI vision request - provider selection {"requested_provider":"groq","user_id":1,"options":{"provider":"groq","model":"qwen/qwen3.6-27b"}}
Groq: Analyzing image {"model":"qwen/qwen3.6-27b", ...}
```

So the panel does not merely pick a different adapter — it bills a provider the installation has not configured for this capability.

In every one of the five runs, `attempts` contained exactly one entry: the winner, verdict `quality_ok`. For the image-only PDF, Tika ran first and lost the quality gate; the panel shows no trace of it.

---

## Root cause

`ExtractionProbeService::testFile()` calls `$this->fileProcessor->extractText($relative, $ext)` (`backend/src/Plug/Extraction/ExtractionProbeService.php:99`) without the user id, so the parameter defaults to `null`. Two branches then diverge from the real upload path:

- **Audio.** `FileProcessor::extractFromAudio()` (`backend/src/Service/File/FileProcessor.php:1000`) evaluates `AiFacade::hasConfiguredSttProvider(null)` as false and takes the local-first branch, while a real upload carries a user id and takes the cloud-first branch.
- **Vision.** `AiFacade::analyzeImage()` (`backend/src/AI/Service/AiFacade.php:840`) only consults the configured `DEFAULTMODEL.PIC2TEXT` row when `$userId !== null`. With `null` that block is skipped entirely — including the **global** row — and resolution falls through to the legacy `getDefaultProvider(null, 'vision')` chain (`backend/src/Service/ModelConfigService.php:85`), which is why the request lands on OpenAI with `model: null`.

**Collapsed attempt list.** The built-in cascade is executed as a single `extractText()` call and appended as one row (`ExtractionProbeService.php:101-105`), so `attempts` can only ever contain individual entries for the extra (non-built-in) adapters. Every built-in step — Tika losing the quality gate, the office→PDF conversion, the vision fallback — is invisible.

**Image path.** Chat images do not go through `FileProcessor` at all; `MessagePreProcessor::processImageWithVision()` calls `AiFacade::analyzeImage()` directly. The panel therefore reports on a code path that image uploads never take (see #1792).

## Steps to reproduce

1. Ensure the account has a `SOUND2TEXT` model configured (a cloud STT provider).
2. `/admin/setup` → tab **Extraction** → **Test upload** with any speech audio file → panel names `whisper_local`.
3. Upload the same file in a chat.
4. `docker compose logs --since 60s backend | grep -a 'Chat file extracted'` → `"strategy":"whisper_api"`.
5. Repeat with an image and compare the `AI vision request - provider selection` lines of both runs.

---

## Notes

This matters more than the wording suggests: the winning adapter is not persisted anywhere. `BFILES` stores only the extracted text in `BFILETEXT` and has no column for the extractor (`BPROVIDER` records the generating provider of AI-generated media, not the extractor). The chat upload response carries `extraction_strategy` only on failure, and no SSE event exposes it. An operator without container-log access has this panel and nothing else, so a wrong answer here is not a cosmetic defect — it is the entire observability surface being wrong.

Direction: pass the requesting user's id into `extractText()` so the probe resolves the same branch and the same models as a real upload, and record each built-in step as its own `attempts` entry instead of one collapsed row.

Related: #1786 (same service, verdict mapping), #1792 (the chain the panel is meant to demonstrate has no effect on built-in adapters).

## Verification

- With a cloud STT model configured, the panel and a real chat upload of the same audio file name the same strategy.
- The panel's vision call uses the provider and model configured in `DEFAULTMODEL.PIC2TEXT`.
- Testing an image-only PDF lists at least two attempts: the failed Tika step with its gate reason, then the vision step.

orgaralf:
## Summary
- FileProcessor now walks the **saved** family chain (built-ins and extras) in order. An adapter that is not in the chain does not run.
- `PUT /api/v1/admin/plugs/extraction/chains` rejects an empty family (HTTP 422). The UI blocks the same case. A previously stored empty string falls back to the seeded default so extraction is not bricked.
- Chat image uploads go through FileProcessor (same as documents/audio), so `image = vision` produces a `FileProcessor` / `Chat file extracted` log.
- The Extraction **Test upload** panel passes the admin's user id into `extractText()` and lists every chain step in `attempts` (Tika losing the quality gate, then vision, etc.). Audio/vision therefore use the same STT / PIC2TEXT models as a real upload.

Fixes #1792
Fixes #1793

## Test plan
- [x] `make -C backend lint`
- [x] `make -C backend test`
- [x] `make -C frontend lint`
- [x] `make -C frontend test`
- [ ] `/admin/setup?tab=extraction`: set `document` to a single adapter, upload a text-layer PDF, confirm logs name that adapter (not always `tika`).
- [ ] Empty a family and Save — rejected; toast explains each file type needs an adapter.
- [ ] Upload a PNG in chat — `FileProcessor: Starting extraction` / `Chat file extracted` with `kind: image`.
- [ ] With cloud STT configured, set `audio = whisper_local,stt_cloud`, upload MP3 — strategy is `whisper_local`.
- [ ] Test upload of the same MP3 / PNG names the same strategy and provider as the chat upload; image-only PDF lists Tika then vision.

### DIFF
--- a/backend/src/Controller/AdminPlugsExtractionController.php
+++ b/backend/src/Controller/AdminPlugsExtractionController.php
@@ -126,7 +126,7 @@ final class AdminPlugsExtractionController extends AbstractController
         )
     )]
     #[OA\Response(response: 403, description: 'Admin access required')]
-    #[OA\Response(response: 422, description: 'Unknown family or adapter key')]
+    #[OA\Response(response: 422, description: 'Unknown family, adapter key, or empty chain')]
     public function saveChains(Request $request, #[CurrentUser] ?User $user): JsonResponse
     {
         if ($resp = $this->requireAdmin($user)) {
@@ -148,7 +148,7 @@ final class AdminPlugsExtractionController extends AbstractController
     #[Route('/test', name: 'admin_plugs_extraction_test', methods: ['POST'])]
     #[OA\Post(
         path: '/api/v1/admin/plugs/extraction/test',
-        summary: 'Probe every configured extra extractor then the built-in path',
+        summary: 'Probe the configured extraction chain with the same path as a real upload',
         security: [['Bearer' => []]],
         tags: ['Admin Plugs Extraction']
     )]
@@ -206,6 +206,7 @@ final class AdminPlugsExtractionController extends AbstractController
         return $this->json($this->extractionAdmin->testFile(
             $file->getPathname(),
             $file->getClientOriginalName() ?: $file->getFilename(),
+            $user?->getId(),
         ));
     }
 
--- a/backend/src/Controller/MessageController.php
+++ b/backend/src/Controller/MessageController.php
@@ -736,20 +736,17 @@ class MessageController extends AbstractController
             $this->em->flush();
 
             // Extract text synchronously for files the chat handler will analyze
-            // immediately (audio, documents). Image extraction is left to the
-            // vision-model pipeline, which reads the file directly at analyze
-            // time. Doing extraction here closes the race documented in
-            // issue #729: the stream used to run FileAnalysisHandler before
-            // the async/deferred extraction had populated BFILETEXT, yielding
-            // a false "Document text extraction failed" error on the user's
-            // first send. By the time uploadFileForChat returns, the file is
-            // either `extracted` (text available) or `error` (extraction
-            // failed) — never `uploaded` with empty text.
+            // immediately (audio, documents, images). Images used to skip this
+            // and go through MessagePreProcessor::processImageWithVision(), which
+            // ignored the extraction chain (issue #1792). Empty OCR on a photo
+            // is legitimate — only audio/documents treat empty text as failure.
             $extractMeta = [];
             $isAudio = in_array($fileExtension, MessagePreProcessor::AUDIO_EXTENSIONS, true);
             $isDocument = in_array($fileExtension, MessagePreProcessor::DOCUMENT_EXTENSIONS, true);
+            $isImage = in_array($fileExtension, MessagePreProcessor::IMAGE_EXTENSIONS, true);
 
-            if ($isAudio || $isDocument) {
+            if ($isAudio || $isDocument || $isImage) {
+                $kind = $isAudio ? 'audio' : ($isImage ? 'image' : 'document');
                 $messageFile->setStatus('extracting');
                 $this->em->flush();
 
@@ -761,13 +758,14 @@ class MessageController extends AbstractController
                     );
 
                     $messageFile->setFileText($extractedText);
-                    $messageFile->setStatus(empty(trim($extractedText)) ? 'error' : 'extracted');
+                    $emptyIsFailure = !$isImage;
+                    $messageFile->setStatus($emptyIsFailure && empty(trim($extractedText)) ? 'error' : 'extracted');
                     $this->em->flush();
 
                     $this->logger->info('Chat file extracted', [
                         'user_id' => $user->getId(),
                         'file_id' => $messageFile->getId(),
-                        'kind' => $isAudio ? 'audio' : 'document',
+                        'kind' => $kind,
                         'text_length' => strlen($extractedText),
                         'strategy' => $extractMeta['strategy'] ?? 'unknown',
                     ]);
@@ -775,7 +773,7 @@ class MessageController extends AbstractController
                     $this->logger->error('Chat file extraction failed', [
                         'user_id' => $user->getId(),
                         'file_id' => $messageFile->getId(),
-                        'kind' => $isAudio ? 'audio' : 'document',
+                        'kind' => $kind,
                         'error' => $e->getMessage(),
                     ]);
 
--- a/backend/src/Plug/Extraction/Adapter/DoclingExtractor.php
+++ b/backend/src/Plug/Extraction/Adapter/DoclingExtractor.php
@@ -12,7 +12,7 @@ use App\Plug\PlugDescriptor;
 use App\Plug\PlugHealth;
 
 /**
- * docling-serve adapter. Lands in FileProcessor via extraExtractorKeys().
+ * docling-serve adapter. Lands in FileProcessor via the configured family chain.
  *
  * @internal
  */
--- a/backend/src/Plug/Extraction/ExtractionAdminService.php
+++ b/backend/src/Plug/Extraction/ExtractionAdminService.php
@@ -75,9 +75,9 @@ final readonly class ExtractionAdminService
      *     markdown: bool
      * }
      */
-    public function testFile(string $absolutePath, string $originalName): array
+    public function testFile(string $absolutePath, string $originalName, ?int $userId = null): array
     {
-        return $this->probe->testFile($absolutePath, $originalName);
+        return $this->probe->testFile($absolutePath, $originalName, $userId);
     }
 
     /**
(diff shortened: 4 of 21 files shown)


orgaralf:
Copilot review follow-up:

**Accepted**
- Empty OCR on a photo (`status=extracted`, empty `fileText`) no longer runs Vision a second time in `MessagePreProcessor`.
- `setChains()` validates every family before any `setValue()`, so a later empty family no longer leaves a partial save.
- Office→PDF Tika/vision results are gated as PDF (`qualityApplyTo=['pdf']`), so low-quality converted text falls through instead of winning as `not_applicable`.

**Declined**
- Removing Tika’s / `pdf_vision`’s Office→PDF prep when `office_convert` is not in the chain. `office_convert` is the legacy `.doc`→`.docx` rewrite. Office→PDF inside `tika` / `pdf_vision` is how those extractors read a DOCX, not that chain step. Dropping it would break a `tika`-only or `pdf_vision`-only document chain.

PR #1823, opened 2026-09-11
feat(admin): switch Assistants, workflows and Export & import from System config

FExB17:
## Problem

Four shipped features are seeded off and have no way to turn them on:

| Flag | Feature | Surface today |
| ---- | ------- | ------------- |
| `AGENTS.ENABLED` | Assistants gallery, builder, pin-to-chat, assistant shares | none |
| `AGENTS.ROUTABLE_ENABLED` | routable assistant takes part in classification | none |
| `WORKFLOWS.BUILDER_ENABLED` | Steps editor on a Saved Task + webhook trigger | none |
| `BUNDLE.ENABLED` | `synaplan-bundle.v1` Export & import | none |

No admin field, no console command, no documented path — the only way to switch them on is an `UPDATE BCONFIG` by hand, in production against the Galera cluster. The Agent Builder plan's own rollout rule ("seed the flag on for new installs after S4") never happened, S1–S6 are all delivered, and the feature is the Foundation item an integration partner is being shown.

## Change

- The four flags are fields in `SystemConfigService::getSchema()` like every other database-backed flag: **Routing → AI assistants**, **Routing → Saved Tasks**, **Interface → Export & import**. No frontend change — tabs and sections render from the schema.
- All four resolve the per-user row before the global one, so writing the global switch now drops the acting administrator's own override. Saved tasks, MCP client, multitask routing and async media already did this as four copies of the same block; that dispatch is now one table plus one helper.
- `docs/CONFIGURATION.md` gets a section for the three groups, `docs/ADMIN.md` a short "how to turn this on" table. The publishing walkthrough said "AI → Instructions", which is the pre-Assistants path — corrected.

## Verification

- `make -C backend lint`, `make -C backend phpstan` (whole project incl. `tests/`), `make -C backend test` — 6214 tests green.
- New tests: each flag reaches its own BCONFIG group row, enabling assistants clears the acting admin's override, a write without an acting user deletes nothing, the new tab sections list the expected fields.
- Added a structural test that every schema field is rendered by an existing tab section and every listed field exists. A field pointing at a missing section looks configured and is invisible — which is close to the state these flags were in.
- Playwright skipped: backend-only, no E2E or Vitest spec touches the admin config schema, and no UI contract changed (the response only gains fields).

## Mobile impact

`backend-only` — `node scripts/mobile-impact.mjs --base origin/main --head HEAD`. Backend plus docs, no `frontend/**`.


orgaralf (approved):
Thanks!

PR #1845, opened 2026-09-13
feat(backend): enable MODULES.GATE_TIKA by default for new installs

orgaralf:
## Summary
- Intermezzo S4 FM21, first module: new installs seed `MODULES.GATE_TIKA=1`. Other gates stay off. Insert-if-missing only — no migration for existing installs.
- Admin System config shows the shipped default as on for Tika.
- Rollout log in `STATUS.md` points at the #1843 `@minimal` E2E path and CI run (matrix PR; this change does not depend on those files).

Replaces stacked #1844, which never started CI (`pull_request` only targets `main`/`develop`).

## Test plan
- [x] `make -C backend lint` + `phpstan`
- [x] `make -C backend test` (6434 tests)
- [ ] CI on this PR (targets `main`)

PR #1853, opened 2026-09-13
feat(backend): enable MODULES.GATE_THEHIVE by default for new installs

orgaralf:
## Summary
- Intermezzo S4 FM21: new installs seed this module gate on. Insert-if-missing only — no migration for existing installs.
- Stacked on the earlier FM21 drafts (#1845 onward). Diff vs `main` includes those commits until they merge.

## Test plan
- [x] `ModuleGateSeederTest`
- [ ] CI on this PR (targets `main`)


orgaralf:
Superseded by https://github.com/metadist/synaplan/pull/1856 — that branch already contains this commit plus the rest of the FM21 stack. Closing to collapse the one-module-per-PR series into a single review.

PR #1858, opened 2026-09-13
docs(planning): close Intermezzo in the live roadmap

orgaralf:
## Summary
Closing off done work

## Changes
The plan still listed the two bugfixes and Intermezzo S4 as next work after both had landed. Record the Wave 5 remainder and keep the UX / use-case docs without dropping the FeatureModule rule.

## Verification
- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated


orgaralf:
Addressed the Copilot review on this PR (`144c79a`).

**Inline**
- Feature-modules plan header now says S1–S4 closed (S5 optional), matching `STATUS.md`.
- Compute §7: only row 5 stays product-owner work; row 4 is already recorded.
- Case 4 utterance is now Turkish (`Taşınmayı nasıl bildiririm?`).

**Suppressed**
- `AGENTS.md` planning rule is scoped to sprint files opened after 2026-09-13. Tools S5 also lists the five UX-contract §6 exit bullets so the live file is not left non-compliant.
- Q7 is **Partial** (#1821 did tool + arguments + permission re-check; tool *version* is follow-up).
- Office Phase T/A/B and IAM-UX really are on `main` (#1685 / #1726). The track STATUS ledgers were still labelled pre-merge; they (and the live-roadmap IAM row) now cite those PRs.

PR #1916, opened 2026-09-15
fix(frontend): save Android downloads instead of opening the share sheet

FExB17:
## Summary
- Native Android **Download** was implemented as `Share.share()`, so tapping it opened the system send sheet (WhatsApp, Telegram, Quick Share) instead of saving the file.
- Android now writes the blob into the public **Documents** folder via `@capacitor/filesystem` (numbered suffix if the name already exists) and does not open the share sheet.
- **iOS is unchanged:** the binary does not enable file sharing, so Cache + share sheet ("Save to Files") remains the supported save path.

Fixes the behaviour described in metadist/synaplan-apps#32.

Classification: `store-required` under the mobile-impact policy (`frontend/src/services/api/native*.ts` is a native seam). No new plugins or Android permissions; Android 13+ needs none for app-created Documents files.

## Test plan
- [ ] Android: Files → download an item (e.g. `recording.webm`) — no share sheet; file appears under Documents in the system Files app
- [ ] Android: download the same name twice — second file is `name (1).ext`
- [ ] iOS: download still opens the share sheet; Save to Files still works
- [ ] Web: browser download is unchanged
- [x] `npx vitest run tests/unit/api/nativeDownload.spec.ts`


orgaralf (approved):
Useful!

PR #1925, opened 2026-09-15
Fix/api session summary language

orgaralf:
## Summary
Fix for the nextcloud integration

## Changes
Speaking the right language for summaries

## Verification
- [ ] Not tested (explain)
- [X] Manual
- [ ] Tests added/updated


orgaralf:
Addressed the Copilot review:

- **Heredoc parse error** — closer is now on its own line; the output-language directive is concatenated after the string.
- **German `bitte`** — added as a detection anchor (aligned with `MessageClassifier`).
- **Tie / single-hit English** — detection now requires a strict lead over the runner-up, so `qué` vs `is` falls back to the account locale instead of picking English by array order.
- **English-detection test** — uses a German account locale so the pin is proven from the excerpt, plus coverage for `bitte` and the Spanish tie fallback.

Local Docker was down, so `php -l` is what ran here; CI is the remaining gate.

PR #1940, opened 2026-09-16
chore(renovate): group gomod updates into one PR

FExB17:
## Problem

The `gomod` manager matched no group rule. Every group in `renovate.json5` scopes itself with `matchManagers` (`composer`, `npm`, `dockerfile`, `docker-compose`) or an explicit package list, so each Go update in `sidecars/synaplan-compute/go.mod` opens an individual PR.

## Change

Add a `Go Sidecar Dependencies` group for `matchManagers: ["gomod"]`, `minor`/`patch`, in the same shape as the existing manager groups. It is inserted before the Docker rules; no rule after it matches `gomod`, so the last-match-wins ordering that the surrounding comments depend on (#580, #1725) is unaffected.

No existing package rule is modified.

## Major splitting stays unchanged

`:separateMultipleMajorReleases` is deliberately kept. One PR per intervening major is what allows stepping a dependency up a version at a time when the upgrade has a compatibility chain, which is worth more than a shorter PR list.

The known cost is recorded in a comment above `extends`: a dependency that lagged several majors arrives as several PRs at once. `github.com/docker/docker` sits at `v23.0.8+incompatible` (`sidecars/synaplan-compute/go.mod:6`) and produced four — #1903 (v25), #1904 (v26), #1918 (v27), #1919 (v28) — which together with the six working group branches reached `prConcurrentLimit: 10` and stalled the rest of the queue until triaged. Merging the newest major autocloses the lower branches.

Two constraints found while checking whether this could be automated, both recorded in the same comment:

- Renovate has no option to close superseded major PRs. A major PR is only autoclosed once its update becomes irrelevant (e.g. a higher major is merged); while v23 is installed, v25–v27 are all considered wanted.
- `separateMultipleMajor` cannot share a `packageRule` with `matchUpdateTypes` — the validator rejects it with `packageRules cannot combine both matchUpdateTypes and separateMultipleMajor`. It has to stay in `extends`.

## Verification

```
$ npx renovate-config-validator renovate.json5
 INFO: Validating renovate.json5
 INFO: Config validated successfully against 1 file(s)
```

Run with the current Renovate release. An older cached validator (37.440.7) rejects `packageRules[0].minimumReleaseAgeBehaviour`; that option is pre-existing (#1486) and valid in current Renovate, not introduced here.

`renovate.json5` is bot configuration only — no application code, so the backend/frontend CI gate does not cover it, and the validator above is the applicable check.

Note: the branch name still carries a `-drop-multi-major` suffix from the first revision of this PR, which dropped the preset. That approach was reverted; only the `gomod` group remains.


orgaralf (approved):
Thanks

PR #1985, opened 2026-09-17
test(e2e): fix the shared arrange wait and share the stream terminal race

FExB17:
## What this changes

Two shared code paths carry most of the recorded CI flakiness. Both are fixed here; no product code is touched.

### 1. `openApp()` — the arrange step of almost every spec

`helpers/auth.ts:127` waited for the initial `GET /api/v1/chats` with `TIMEOUTS.STANDARD` (10s). `page.waitForResponse: Timeout 10000ms exceeded` was the most frequent first-attempt signature in the available Playwright artifacts: 9 of 17 flake instances across 52 runs, spread over `navigation.spec.ts` (4x), `multitask.spec.ts`, `memories.spec.ts`, `mcp-config.spec.ts` and `chats-archive.spec.ts` — different specs, one shared wait.

This is the first request of the first test in a worker, against a backend sharing four runner cores with MariaDB, Qdrant, the stubs and four browsers. It now uses `VERY_LONG`, which costs nothing on a warm stack because `waitForResponse` returns as soon as the response lands.

The predicate also now requires `res.ok()`. Without it a 401 satisfied the wait, the app then refreshed and re-fetched, and the helper returned into exactly the `ensureValidActiveChat()` race it exists to prevent.

### 2. `login()` — an error message that named the wrong subsystem

`helpers/auth.ts:180` replaced the Playwright error with `Login failed: current URL is <url>`. The sign-in form is a guest-only route, so the `http://localhost:8001/` seen in CI logs means the sign-in **succeeded** and only the composer had not rendered. The message claimed the opposite. Flakes carrying this signature were booked against `admin-impersonation-chat.spec.ts`, which is not where the wait lives.

It now distinguishes the two cases and keeps the original error as `cause`, so the failing selector survives in the report.

### 3. The done-vs-error race existed once, inside `ChatHelper`

Every surface that waited on a stream without it could only observe success: a failed stream burned the full timeout and reported `Test timeout of 60000ms exceeded`, which names no step, and then usually passed on retry. Moved to `helpers/stream.ts` (`assertStreamSucceeded`) and adopted by `multitask.spec.ts` (three call sites).

Verified against a live error on the dev stack: fails in 4.8s naming the notice sentence instead of timing out.

```
Error: Assistant message ended in error state: Something went wrong while answering.
       Nothing new was saved as the final reply.
  at assertStreamSucceeded (helpers/stream.ts:33:11)
  at ChatHelper.waitForAnswer (helpers/chat.ts:57:5)
```

### 4. Smaller items in the same class

- `widget.spec.ts:193` and `chat-again.spec.ts:39,68`: dropped the `attached` -> `scrollIntoViewIfNeeded` -> `visible` sequence. `ChatHelper.waitForAnswer` removed it because `scrollIntoViewIfNeeded` snapshots the element handle and throws `Element is not attached to the DOM` when the bubble re-renders mid-call. These were the last `@ci` callers still carrying it; the widget one runs in both the chromium and firefox jobs.
- `incognito.spec.ts:18`, `memories.spec.ts:69`, `chat-share.spec.ts:8`: wall-clock headroom. These three call `waitForAnswer` on the 60s default, so the step waits above could not be reached — the test died on the wall clock and reported no failing step.
- `selectors.ts`: removed `messageTopicError`, an unused duplicate of `chatError` (both `[data-testid="chat-error-notice"]`), and centralized `chat-error-body`, which was inlined in the helper.

## Verification

Local gate, all green: `eslint` (0 errors), `vue-tsc -b --noEmit`, Vitest 2006/2006.

E2E: the `@ci` chromium suite was run twice against the local dev stack, once on this branch and once on `048acd02e` (`origin/main` at branch point).

| | failures | passed |
| --- | --- | --- |
| `origin/main` 048acd02e | 40 | 93 |
| this branch | 34 | 99 |

The failure sets were diffed by test name. **`branch \ main` is empty — no new regressions.** Six tests fail on main and pass here (5 in `guest-chat.spec.ts`, 1 in `admin-panel.spec.ts`), but two runs of this branch alone differed by 7 tests, so those six are inside this environment's noise band and are not claimed as fixes.

The remaining failures in both runs are the dev stack lacking services the `@ci` suite expects from `docker-compose.test.yml`: `ECONNREFUSED ::1:8026` (MailHog), `ECONNREFUSED ::1:3999` (WhatsApp stub), `ENOTFOUND host.docker.internal` (Keycloak), Stripe `Invalid signature`, and `chat-error-notice` from the absent chat provider. CI runs the suite against the test stack on `:8001`, so its result is the authoritative one — hence draft.

## Not in this PR

- **Impersonation.** The remaining race is product-side: four independent token-refresh pools (`httpClient.ts:246`, `chatApi.ts:114`, `authService.ts:99`, `apiService.ts:53`) can each have a `POST /auth/refresh` on the wire during the cookie swap, while `stores/auth.ts:476` awaits only `getInFlightRefresh()` (the httpClient pool). A late `Set-Cookie` clobbers the impersonation cookie, `/auth/me` sees no `impersonator`, and the banner never mounts. Consolidating those pools is a change to auth transport and belongs in its own PR.
- **Widget error race.** `ChatWidget.vue` renders no per-message error terminal state, so `waitForWidgetAnswer` cannot race one without a new `data-testid` in product code. Its three callers are `@noci`.
- **Flake reporting.** `e2e-flaky-notify` has no persistence and no normalization by run count, and does not separate CI-infrastructure failures (5 of 15 sampled hard E2E failures were `packages.microsoft.com 403` or `quay.io` timeouts). Playwright 1.63 supports `failOnFlakyTests`, unused here. Workflow changes need their own review.


orgaralf (approved):
thank you

PR #1990, opened 2026-09-17
ci: bound every job runtime

FExB17:
## Summary

- Every job in `ci.yml` (23) and the visual-baseline recorder now sets `timeout-minutes` immediately after `runs-on`, which is GitHub's documented job key.
- GitHub's default is 360 minutes. A hung step was free to occupy a runner that long and still report success: job `103217995854` (run `34584937152`) took 270 minutes, 269 of them in one step, then ran its tests in 24 seconds.
- Caps sit well above observed medians (E2E 3.8 min → 20; Docker build 1.5 min → 30; arm64 publish 40) so ordinary variance still passes and only a hang is cut off. Playwright stays on `install --with-deps` / `install-deps`.

## Scope

CI configuration only. Playwright OS packages are unchanged: `--with-deps` is Playwright's documented CI path, and a census of 486 recent apt steps (median 24 s, one 403) did not justify dropping them.

## Test plan

- [ ] CI on this PR is green (timeouts must not trip healthy jobs)
- [ ] `actionlint` has no new findings beyond the existing shellcheck nits on `main`
- [x] YAML parses; every job has `timeout-minutes` immediately after `runs-on`

## Mobile impact

`backend-only` — workflows only, no `frontend/**` application code.


FExB17:
Scope reduced to job timeouts only.

`--with-deps` / `install-deps` stay: that is Playwright's documented CI path. Apt-mirror failures in the recent window were rare (one 403 in 486 steps, median 24 s); dropping apt would trade that for a latent runner-image drift risk. History was rewritten to a single commit on `ci/bound-job-runtimes`.

FExB17:
Superseded by #1991. This PR closed when the branch was renamed after the apt change was dropped.

PR #1989, opened 2026-09-17
fix(catalog): retire TrustedTokens DeepSeek V4 Flash 0731 and V4 Pro

FExB17:
## Summary
- TrustedTokens dropped `deepseek-ai/DeepSeek-V4-Flash-0731` and `deepseek-ai/DeepSeek-V4-Pro-0813`; the hourly health check confirmed both Gone at 2026-09-17 12:16 UTC.
- Record retirements for BID 336 and 337 (`active = selectable = 0`); ModelRetirementSeeder stamps `BRETIREDON` / `BSUCCESSORID` on deploy. No migration.
- Re-point BID 335's successor from Flash-0731 to GLM-5.3-Flash so a retirement chain cannot land on another dead model. Flash-0731 → GLM-5.3-Flash (same price); V4 Pro → GLM-5.3.

## Test plan
- [x] `make -C backend lint`
- [x] `make -C backend phpstan`
- [x] `make -C backend test` (6825 tests)
- [ ] After merge: `app:seed` on deploy applies the retirement; Admin → Model status should stop listing these two as incidents.

Issue #1515, opened 2026-08-19
Models: retirement data model


FExB17:
## Summary

Solved in #1532. Retiring a model is one reviewable `ModelCatalog::RETIREMENTS` entry. `ModelRetirementSeeder` stamps `BRETIREDON` / `BSUCCESSORID` and deactivates the row on every deploy. `ModelCatalogRetirementTest` fails a catalog drop that has no retirement record.

---

## Problem / Motivation (original)

Model identity was a bare `BMODELS.BID` with no retirement state and no successor. Every provider shutdown needed a hand-written migration. Three of five such migrations existed only to clean up rows dropped from `ModelCatalog` while left `BACTIVE = 1, BSELECTABLE = 1` on existing installs.

That maintenance fire is gone. New retirements (for example Imagen 4 in #1591) are registry entries, not SQL.

---

## What shipped

- `BMODELS.BRETIREDON` + `BMODELS.BSUCCESSORID` (`Version20260820150000`), mapped on `App\Entity\Model`
- `ModelCatalog::RETIREMENTS` — date, successor catalog key or explicit `null`, reason, `providerId` guard
- `ModelRetirementSeeder` — idempotent, `BPROVID`-guarded, runs from `app:seed`
- Request path treats `BACTIVE = 0` as unusable (`ModelConfigService::isModelUsable` / `getDefaultModel` / `resolveUsableModelId`) and falls back to a live capability default
- Snapshot gate: `tests/Unit/Model/ModelCatalogRetirementTest.php`

---

## Acceptance criteria

- [x] `BMODELS` carries a retirement state and a successor reference, readable by application code
- [x] Retiring a model is one catalog entry, applied idempotently on deploy, safe to run repeatedly
- [x] A model dropped from `ModelCatalog` without a retirement record cannot reach a release
- [x] The main request path does not hand a deactivated row to a provider when any live model for that capability exists

---

## Notes

related: #1512, #1513, #1532, #1591

Items that appeared as extra checkboxes (follow `BSUCCESSORID` instead of the capability default, user-facing errors for an explicit pick, admin retirement UI, rewriting every stored BID including `BMESSAGEMETA.MODEL_ID`) are not required for this issue. Message model ids are history. Embedding pointers must not auto-repoint (#948). `successor: null` is deliberate for VECTORIZE and for models with no replacement.

`docs/PRICING_MAINTENANCE.md` still mentions successor consumption as a follow-up; that is optional polish, not an open hole in the retirement data model.

[...]

PR #1992, opened 2026-09-17
refactor(catalog): make a model retirement one registry entry

FExB17:
Retiring a model touched six places, and the last retirement ([#1989](https://github.com/metadist/synaplan/pull/1989)) touched exactly those: the `RETIREMENTS` entry, the catalog row's `active`/`selectable`, the provider class docblock, `README.md`, `docs/CONFIGURATION.md` and the maintenance log. Two of those are gone, a third is now caught by a test, and the entry itself is written for you.

## The row flags are derived, not repeated

`ModelCatalog::rows()` forces `active`/`selectable` to `0` for every BID in `RETIREMENTS`, so a row you keep in the catalog is switched off by the entry alone. It was a second manual step whose only guard was `ModelCatalogRetirementTest`; forgetting it shipped fresh installs a row they would offer, select and bill for after the provider had switched it off.

No behaviour changes today: all eight retired rows still in the catalog already carried `0/0`, and `fingerprint()` does not cover those columns, so no stored fingerprint moves and no row falls out of `ModelSeeder`'s update path.

## A retired id may not be a fallback — two live bugs

`RetiredModelReferenceTest` scans `backend/src/` for a retired provider id in a *default* position (`??`, `?:`, `return`). It found two:

| Site | Fallback | Retired |
| --- | --- | --- |
| `GoogleProvider::generateImage()` | `$options['model'] ?? 'imagen-4.0-generate-001'` | 2026-08-17, Google hard-shut the endpoint |
| `OpenAIProvider::pickResponsesModel()` | `return 'gpt-4.1'` for every `gpt-image-*` call | 2026-07-27, left the catalog |

The Google one is reachable: `SyncMediaJobGenerator::generateImage()` only sets `model` when the job carries one, so an image job without a model called a 404 endpoint for a month. Both now use the successor the registry itself records (`RETIREMENTS[115]` → Nano Banana 2, `RETIREMENTS[30]` → GPT-5.6 Terra).

The guard is deliberately narrow. A retired id is legitimate in a family check (`str_starts_with($model, 'gpt-5')`), in `AnthropicProvider::SUPPORTED_MODELS` (the ids Claude Code sends us) and in an OpenAPI `example:` — 23 such mentions exist and none of them decides where a request goes. Banning them all would have been a lie with a 21-line allow-list.

## The report writes the entry

For every model it confirms Gone, `app:models:check-availability` now prints the paste-ready block: keyed by BID, dated today, `reason` naming the provider, and the closest live row of the same provider and capability as the suggested successor.

Checked against the successors humans picked by hand: identical for BIDs 335, 336 and 231, and a neighbouring tier for 337 and the other two Imagen rows. `null` when the catalog has no sibling, which is the entry that forbids a substitution.

## Prose, honestly

`README.md` and `docs/CONFIGURATION.md` still advertised "Imagen 4" a month after the shutdown, and no test can tell that an editorial summary became a lie — the retired row is named `Imagen 4.0`, the prose says `Imagen 4`, and `DeepSeek V4 Flash` is simultaneously dead at TrustedTokens and live at A2Agent. Both files are corrected here, the README's claim that retirements "ship as seeders plus a migration" is corrected too, and `docs/PRICING_MAINTENANCE.md` now names this as the one part of a retirement that still needs a person.

## Gate

`make -C backend lint`, `make -C backend phpstan`, `make -C backend test` (6831 tests) all green, unfiltered. Frontend untouched; no HTTP or OpenAPI contract change, so Playwright is not required. `scripts/mobile-impact.mjs` classifies the change as **backend-only** (docs as no-app-impact), no new path classes.


FExB17:
Follow-up in `55bc4d1a7`, from auditing how past retirements were actually written.

**The guard missed the shape that has broken before.** Scoping it to `??` / `?:` / `return` left out the one that already cost a hand-fix: thirteen provider classes keep a `DEFAULT_CHAT_MODEL` / `DEFAULT_VISION_MODEL` constant, and `GroqProvider::DEFAULT_VISION_MODEL` still named `meta-llama/llama-4-scout-17b-16e-instruct` when Groq dropped it — corrected by hand in [#1513](https://github.com/metadist/synaplan/pull/1513), caught by nothing. A retired id is now flagged wherever a model is *chosen*: assigned (`=`, `??=`, which covers a `const` and a parameter default), defaulted (`??`, `?:`) or returned. All 18 of those constants currently name live models, so this closes the hole rather than fixing a third bug.

Widening to `=` added no noise: `=>` is its own token, so capability lists and OpenAPI examples stay out, and comparisons have their own tokens, so `$model === 'gpt-5'` is not a choice. The position logic is now pinned on a snippet holding every shape at once — a guard that has never been seen to fire proves nothing by passing.

**`CONTEXT_PRICING` was one more line to remember.** A long-context tier keyed on a model no live row carries is dead config that reads like a live rate. `ModelCatalogRetirementTest` now fails on it. Keyed by `providerId`, so retiring the chat row of a model whose vision row lives on is still fine. Currently clean.

Gate re-run unfiltered: lint, phpstan, `make -C backend test` (6833 tests) green.

FExB17:
Follow-up in `339ea6370`: `BSUCCESSORID` is now read at resolution time.

A stored widget/`DEFAULTMODEL` BID that is retired is rewritten to the catalog successor before the request goes out. `successor: null` still falls through to the capability default.

Checked on this machine against the seeded DB, incognito stream as `demo@synaplan.com`:

- override BID **9** (Groq Llama 3.3, retired) → log `configured_model_id: 9, successor_model_id: 76`, stream generated with Groq `gpt-oss-120b`
- override BID **321** (xAI grok-stt, no successor) → no successor log, fallthrough to BID **249** (Sonnet 5)

Admin UI still does not show retirement state — that remains the leftover from #1515.

PR #2038, opened 2026-09-20
fix(planner): route explicit script-execution requests to code_run

orgaralf:
## Root cause

Rule 3 of `tools:plan` routes **every** image generate/edit to `image_generation` — including *"write a python script and apply a headline across the attached image"*, which the planner sent to the AI image model (PIC2PIC) instead of `code_run`. `code_run` had **zero** arbitration text in the 31 KB prompt (18 `image_generation` mentions vs 0 `code_run`).

Traced on a live turn (local message 39): sorter voted `mediamaker`/`BMULTI=true` (correct, v5.0.0 works) → planner ran (~1.5 s, silent) → single `image_generation` node → legacy delegation → image2image. Every layer worked as written; the prompt gave the model no way to choose `code_run`.

## Fix

- New **rule 3a**: script/program/code the user asked to be RUN → `code_run`, wins over rules 3 and 6; rule 3 defers to it. Script text *only* ("show me the code") stays `chat`.
- `PromptCatalog::planPrompt()` for fresh installs + idempotent anchor-based migration `Version20260921010000` for existing DBs (same surgical pattern as `Version20260920210000`: global row only, skips customized prompts, Galera-safe single UPDATE). Verified migration output is byte-identical to the catalog change, down() round-trips.
- Golden planner-prompt snapshot re-recorded (diff is exactly the rule).
- Two `plan-eval` corpus cases + `PromptCatalogTest` assertion.

## Plan-eval (groq gpt-oss-120b, live model)

| run | result |
|---|---|
| new cases, original prompt | 0/2 (bug reproduced: `chat` + `image_generation`, no `code_run`) |
| `script_headline_on_attached_image`, fixed prompt | **4/4 PASS** → `code_run` |
| `script_csv_totals`, fixed prompt | 6/8 PASS → `code_run` (2 shape-only flakes: bare-array output, routing correct in all well-formed outputs) |
| full corpus, original prompt | 33 passed, 6 failed (4 pre-existing, model-stable) |
| full corpus, fixed prompt | 34 passed, 5 failed (**same 4 pre-existing**, no regressions) |

## Files

- `backend/src/Prompt/PromptCatalog.php`
- `backend/migrations/Version20260921010000.php`
- `backend/tests/Eval/plan_eval_corpus.json`
- `backend/tests/Unit/PromptCatalogTest.php`
- `backend/tests/Characterization/__snapshots__/planner_system_prompt.txt`


orgaralf:
Addressed both review findings in ae9bb61cb:

1. **Migration test** — added `Version20260921010000Test` (up/down surgery, idempotency, customized-prompt skip, byte-identity of the migrated real catalog prompt incl. single-insertion check). 5/5 green locally.
2. **`inputFileIds`** — confirmed against `CodeRunRunner::run()`: the planner can't know numeric ids and the runner already falls back to the message's attachments when the field is omitted. Rule 3a and the capability note now say to omit it and never invent ids, instead of asking the planner to fill it. Validator doesn't require the param; snapshot re-recorded. Re-ran eval after the reword: routing to `code_run` correct in all well-formed outputs (remaining flakes are groq shape-only: bare-array/malformed tasks despite structured output being on — provider-side, pre-existing).

PR #2083, opened 2026-09-21
fix(iam): leave keeps shares and owned chats show their grants

### LINKED ISSUE #2054: IAM: Private pill claims exclusive visibility on org-shared chats
## Problem

An owner-owned conversation that is shared with everyone in the organization still carries the **Private** pill (lock icon, tooltip *Your own chat — only you can see it*) in Recent Conversations, and the **Private** filter still counts it.

---

## Expected

The owner's history row must not claim exclusive visibility while an IAM grant or a public link is active. Who else can see the chat is answerable on the row (U7), not only inside the Share dialog. Filter labels that mean *chats I own* must not use the word **Private** if that word also means *nobody else can see this*.

## Actual

Sharing works. Recipients find the same conversation under Incoming chats with the **Everyone** pill. On the owner's side the list still treats the chat as private:

1. Recent Conversations shows the **Private** pill with a lock icon. Hover copy is `iam.incoming.pill.privateTitle`: *Your own chat — only you can see it* (`frontend/src/i18n/locales/en/admin.json`; same claim in de/es/fr/tr).
2. The **Private** filter includes the chat; **Group** stays `0` for the owner.
3. The Share dialog for the same chat lists **Everyone in this organization** / Can use, and can also show a public link (*Anyone with the link can view*).

Observed with an owned chat titled *Plan-Ist-Vergleich DokumentenÜbersicht*: Share listed Everyone / Can use plus a public link; the history sheet still showed Private on that row and Group `0`.

Mechanism (ownership vs grants are orthogonal; the row never consults grants):

- `ChatKind` is list provenance, not IAM visibility. `ChatKindPill.vue` documents `private` as *my own conversation* and still renders the lock icon plus `privateTitle`.
- `SidebarV2.vue` `ownChatList` hard-sets `kind: 'private'` on every owned, non-widget chat. `chatKindCounts.private` is `ownChatList.length`; `chatKindCounts.group` is incoming-only. `matchesChatFilter(..., 'private')` is `kind === 'private'` (`frontend/src/utils/chatKind.ts`). `ChatBrowser.vue` maps owned chats the same way (`kind: 'private'`).
- Org-wide share writes a `BSHARES` row (`subjectType` `everyone`) on the same conversation id. It does not change `BCHATS` or the owner's list `kind`. Recipients see that id on `GET /api/v1/me/shared?kind=conversation` with `sharedVia.type === 'everyone'`.
- The row's link icon is `chat.isShared`, which is the public link flag (`ChatController` list payload: `'isShared' => $chat->isPublic()`), not IAM grants. The chats list has no share-grant summary, so an org share has no owner-row affordance at all.

---

## Steps to reproduce

1. As the owner, open a conversation.
2. Share it with **Everyone in this organization**, permission **Can use**.
3. Open **Recent Conversations**.
4. Observe the **Private** pill and tooltip on that row; **Private** still counts it; **Group** is unchanged.
5. As another user in the same organization, open **Incoming chats** and find the same conversation with the **Everyone** pill.

---

## Notes

`ChatKind` (mine vs incoming) and `BSHARES` (who else) are two axes. The bug is the owner's copy and affordance, not the grant.

Direction:

- Stop using **Private**, the lock icon, and *only you can see it* for ownership. Name the filter/pill as *chats I own* (e.g. **Yours**), and drop the exclusive-visibility claim from `privateTitle`.
- On the owner's row, surface an active IAM grant (or at least stop contradicting the Share dialog). That needs a grant summary on the chats list; `isShared` today means public link only.
- Same mapping exists in `ChatBrowser.vue`; keep that surface in the same change.

`iam.dialog.empty` (*Only you can see this.*) is the empty-grant state and is not this bug.

---

## Screenshots/Logs

Owner Share dialog: Shared with **Everyone in this organization**, Can use; public link *Anyone with the link can view*.

Owner Recent Conversations: same chat under **Private**; **Group** count `0`.

### LINKED ISSUE #2058: IAM: Again with / Continue are offered in a shared chat but the stream endpoint is owner-only
## Problem

Message actions that write to the conversation — **Again**, **Again with…**, **Continue** — are offered to a recipient of a shared chat, but the stream endpoint accepts the owner only, so they fail with "Chat not found or access denied".

---

## Expected

A conversation opened through a share offers only the actions the permission covers. **Can view** allows reading; **Can use** adds **Continue as my copy**. Neither regenerates an answer inside the owner's thread, so those controls must be absent — not present and failing.

## Actual

A recipient presses **Again with…** in a chat shared with their group. The turn ends in an error: "Something went wrong while answering. Nothing new was saved as the final reply.", carrying `{status: 'error', error: 'Chat not found or access denied'}`. The accompanying toast is `chat.notFound` ("The chat no longer exists"), which is wrong — the chat exists and is readable in the same view.

`POST /api/v1/messages/stream` checks ownership directly and never consults IAM (`backend/src/Controller/StreamController.php:887-891`):

```php
$chat = $this->em->getRepository(Chat::class)->find((int) $chatId);
if (!$chat || $chat->getUserId() !== $user->getId()) {
    $this->sendSSE('error', ['error' => 'Chat not found or access denied']);
    return;
```

`GET /api/v1/chats/{id}` resolves the same conversation through `AccessGate` to `owner` | `read` | `use` (`backend/src/Controller/ChatController.php:356-370`, `956-987`), which is why the chat opens at all.

The frontend never narrows the action row: `ChatMessage.vue` has no `readonly` / `access` prop (the regenerate button is `ChatMessage.vue:823-832` → `ChatView.handleRegenerate:5041-5062`), unlike the public token view, which does pass `readonly` (`frontend/src/views/SharedChatView.vue:223`). The composer *is* locked for both permissions (`ChatView.vue:736-757`), so the write path is blocked everywhere except these buttons.

What a non-owner currently sees on a message:

| Action | Offered | Result |
| ------ | ------- | ------ |
| Copy | yes | works |
| File download | yes | works while the share exists |
| Report | yes | works |
| Not correct | yes | records feedback on the recipient's account, not the owner's |
| **Again / Again with…** | **yes** | **fails: owner-only stream** |
| **Continue** (truncated answer) | **yes** | **fails: owner-only stream, also checks `message.userId`** |

Observed in the backend log: `StreamController: Received request user_id:4 chat_id:"32" is_again:true model_id:"324"`, where chat 32 belongs to user 2. The owner check returns without logging, so the failure leaves no server-side trace.

---

## Steps to reproduce

1. As user A, share a conversation with a group that user B belongs to, permission **Can use**.
2. As user B, open the conversation from Incoming chats.
3. On an assistant message, open the action menu and pick **Again with…** (any model).
4. Observe the error state and the "Chat not found" toast.

---

## Notes

Two separate defects:

- The action row is not permission-aware. Pass the resolved `access` into `ChatMessage` / `MessagePart` and hide every action that writes to the owner's conversation. **Can use** keeps **Continue as my copy** on the banner (`SharedConversationBanner.vue:29-37`), which is the supported path.
- The error copy is wrong for this case. "The chat no longer exists" describes a deleted chat; this one is visible on screen.

Regenerating inside someone else's conversation should stay owner-only — the fix is to stop offering it, not to widen `StreamController`.

### LINKED ISSUE #2065: IAM: leaving a group keeps everything the leaving member shared with it
## Problem

Leaving a group does not withdraw what the leaving member shared with it. The former member's conversations, folders and assistants stay readable for the remaining members, and nothing on the Leave action says so.

---

## Expected

Leaving a group is an undo action for the membership, and the consequence must be stated where it is taken (U3). Either the grants that member made to that group are withdrawn with the membership, or the Leave confirmation says in one sentence that they remain and where to revoke them.

## Actual

`GroupService::leave()` and `removeMember()` delete the membership row and write an audit entry, and stop there (`backend/src/Service/Iam/GroupService.php:172-192`, `:199-220`). No `BSHARES` row is touched. `ConversationKind::onShareChanged()` is only invoked from grant and revoke, never from a membership change.

Share cleanup exists for two other events only: deleting the group (`deleteBySubjectGroup`, `GroupService.php:124`) and deleting the user (`deleteByGrantedBy`, `UserDeletionService.php:459`). `ShareRepository` has no query for "grants made by this user to this group".

Access is evaluated against **current** membership of the subject group (`AccessGate.php:69-74` → `ShareRepository::findForSubjects`), so as long as the row `BSUBJECTTYPE='group', BSUBJECTID=<group>` exists, every remaining member keeps the granted permission.

Observed: a member created a conversation, shared it with the group, left the group, and the conversation stayed open to the remaining members. Confirmed in the audit trail — `group.member_leave` is recorded, the share row is not removed.

The former member is not locked out of fixing it: owner is always `manage` (`AccessGate.php:59-60`), the share list is read per resource and resolves the group name from `BGROUPS` rather than from membership (`ShareService.php:536-540`), so the group still shows in the chat's Share dialog and can be revoked there. But nothing points there, and the group no longer appears anywhere in that person's own navigation.

---

## Steps to reproduce

1. As user B, a member of group G, create a conversation and share it with G (any permission).
2. As user B, open Account → My groups and leave G.
3. As user A, still a member of G, open Incoming chats.
4. Observe the conversation is still listed and openable.

---

## Notes

No written decision covers this. The IAM sprint files specify share cleanup on user deletion and on resource deletion; group deletion is the only membership-adjacent cleanup that was implemented (`_devextras/planning/2026-archive/20260903_iam/02_sprint_2_sharing_mvp.md` §2.1–2.2). Leaving was never considered, so this is an unspecified gap rather than a deviation from the plan.

Two defensible resolutions, one decision needed:

- **Withdraw on leave.** In `leave()` / `removeMember()`, delete `BSHARES` where `BGRANTEDBY = <user> AND BSUBJECTTYPE = 'group' AND BSUBJECTID = <group>` — narrower than the existing `deleteByGrantedBy`. Matches the expectation that leaving a group ends the relationship in both directions, and needs a sentence on the Leave confirmation naming how many items will be unshared.
- **Keep the grants** and say so: the Leave confirmation names the count and links to where they can be revoked. The grant was made to the group as an entity, and other members may depend on it.

Related: sharing into a group one does not belong to is possible in the first place, because the subject picker lists every group (`ShareService.php:324-346`) and `assertSubjectExists` only checks existence (`:439-442`). That part is covered by #2060.

### LINKED ISSUE #2066: People: Shared with this group always reports nothing, the panel never queries
## Problem

The "Shared with this group" panel on Operate → People → Groups always reports that nothing is shared, because it is a hardcoded empty state that never queries anything.

---

## Expected

Selecting a group shows what is currently shared with it — metadata only (kind, name, owner, permission), consistent with People → Users, which already lists a user's resources without exposing content.

## Actual

The panel renders the empty string unconditionally (`frontend/src/components/people/GroupDetailPanel.vue:92-97`):

```html
<div>
  <h4 class="text-sm font-medium txt-primary mb-2">{{ $t('people.groups.sharedWith') }}</h4>
  <p class="text-sm txt-secondary" data-testid="group-shared-empty">
    {{ $t('people.groups.sharedEmpty') }}
  </p>
</div>
```

There is no fetch: `iamApi.ts` has no call for group shares, and no `GET /api/v1/admin/groups/{id}/shares` endpoint exists. `GET /api/v1/shares` requires `kind` **and** `resource`, so it answers "who is this item shared with", never "what is shared with this subject". `ShareRepository` can delete by group subject (`backend/src/Repository/ShareRepository.php:163-173`) but has no matching read query.

Verified against the database while the panel showed the empty state: `SELECT * FROM BSHARES WHERE BSUBJECTTYPE='group' AND BSUBJECTID=2;` returns a live conversation grant with permission `use`, and members of that group can open the conversation from Incoming chats.

The panel was planned as empty until sharing shipped (`_devextras/planning/2026-archive/20260903_iam/01_sprint_1_groups_core.md:120`, IAM9: "shared with this group — empty until S2") and S4 specified filling it via `describe()` / `ResourceCard` (`04_sprint_4_directory_and_privacy.md:76`). That step was never wired.

---

## Steps to reproduce

1. As any user, share a conversation with a group.
2. As an admin, open Operate → People → Groups and select that group.
3. Observe "Nothing is shared with this group yet." while members of the group can open the conversation.

---

## Notes

This is the surface an administrator uses to answer "what does this group have access to" before removing a member or deleting the group — and deleting a group silently drops every grant made to it (`GroupService.php:124`), with no way to see beforehand what that removes.

Direction: add `GET /api/v1/admin/groups/{id}/shares` returning metadata through the existing `kind->describe()` path (same shape as `AdminUserResourcesController`, which already does this for a user), back it with a `ShareRepository` query on `(BSUBJECTTYPE, BSUBJECTID)`, and bind `GroupDetailPanel` to it. The empty state stays for groups that genuinely have nothing.

### LINKED ISSUE #2067: People: group member field says Search by email but requires an exact address and offers no suggestions
## Problem

The add-member field on Operate → People → Groups is labelled "Search by email" but performs no search: it shows no suggestions while typing and rejects anything that is not a complete, exact email address.

---

## Expected

Adding a person to a group works like adding a person in the Share dialog: typing a fragment of a name or address lists matching people to pick from, and the selected person is added. A partial name is a normal way to look somebody up, not an error.

## Actual

The input has no `@input` handler and no results list (`frontend/src/components/people/GroupDetailPanel.vue:31-37`). Everything happens on submit (`:139-161`):

```js
const result = await adminApi.getUsers(1, 20, query)
const match = result.users.find(
  (user) => (user.email ?? '').toLowerCase() === query.toLowerCase()
)
if (!match) {
  showError(t('people.groups.addMemberNotFound'))
  return
}
await iamApi.setMember(props.group.id, match.id, memberRole.value)
```

Typing a first name and pressing the button fails twice over: the admin search matches on the mail column only (`backend/src/Controller/AdminController.php:141-144`, `u.mail LIKE :search`), and even a hit would be discarded because the client requires an exact string equality with the full address. The resulting error is `people.groups.addMemberNotFound` — "No user matches that email." — while the placeholder `people.groups.addMemberPlaceholder` promises "Search by email".

The infrastructure for the intended behavior already exists and is used one screen away: `SubjectPicker.vue` debounces `GET /api/v1/iam/subjects?q=`, which searches name **and** email and returns `{ type, id, name, email }` per subject (`backend/src/Service/Iam/ShareService.php:296-332`). The group panel does not use it.

Members are also rendered by email only (`GroupDetailPanel.vue:65-67`), because `GroupService::serializeMember()` returns no display name (`backend/src/Service/Iam/GroupService.php:349-357`) although the name is available in `BUSERDETAILS` and is already resolved elsewhere by `ShareService::displayName()` (`ShareService.php:546-556`).

---

## Steps to reproduce

1. As an admin, open Operate → People → Groups and select a group.
2. Type a first name, e.g. the first three letters of a member's name, into the member field.
3. Observe no suggestions appear.
4. Press **Add a person** and observe "No user matches that email."

---

## Notes

Direction: replace the plain input with the existing subject picker restricted to `type === 'user'` (either a prop on `SubjectPicker.vue` or a small shared user combobox extracted from it), and pass the selected subject id straight to `iamApi.setMember` instead of resolving an email against the admin user list. Add the display name to `GroupService::serializeMember()` and render name with email as the secondary line, matching how people are presented in the Share dialog.

orgaralf:
## Summary

- Leaving a group keeps that member's grants. The confirmation says so. Shares are removed only when the person chooses **Leave and stop sharing**. Removing a member does not delete their shares either.
- A group's admin page lists the real grants, and adding a member searches by name or email without the public directory picker.
- Shared chats hide Again, Again with, and Continue. Continue as my copy stays for people who can use the chat. The owner filter reads **Owned**, and the owner's row shows who the chat is shared with.
- Operate and AI infrastructure accordions start closed. A jump link or `?section=` still opens that one section.

Closes #2054, closes #2058, closes #2065, closes #2066, closes #2067.

## Test plan

- [x] Leave a group with a grant: dialog states the share stays; the grant is still there after Leave
- [x] Leave and stop sharing removes only that person's grants to that group
- [x] Admin member search matches a last name and shows first and last name
- [x] A shared chat hides Again / Continue and keeps Continue as my copy
- [x] Owned filter and grant pills on the owner's chat row
- [x] System configuration opens with every section closed; `?section=cloud` opens only that section
- [x] Unit tests (frontend 2104, backend except the local `COMPUTE_ENABLED` runtime-config assertion)
- [x] E2E for Models & keys cards and the Higgsfield bookmark, updated to open a closed panel


orgaralf:
Addressed the five review notes in d6f38c0db:

- Leave and stop sharing now deletes the grants and removes the membership in one transaction. Resource cleanup runs only after that commit.
- Chat list omits grant summaries when sharing is off, and the chat browser hides the pills the same way the sidebar does.
- A shared chat also hides the error retry, not only Again / Continue.
- The member name field and role select use the standard form surface.
- The people-count pill has its own tooltip, so it no longer says the chat was shared with you.

PR #2122, opened 2026-09-23
fix(chat): keep a finished answer when a stale history load lands

FExB17:
## Summary
- The impersonation E2E flake is a real dropped answer, not a timeout. After impersonation the chat view waits for the model catalog before the first history GET, so that request overlaps the TestProvider stream and returns after the answer has already finished.
- The history store only kept the in-flight turn while `isStreaming` was still true. The stale snapshot (user row only) then replaced the finished reply, so `message-done` never appeared. A load that started during the turn now keeps that turn even after it finishes. A load that starts afterwards still adopts the server snapshot.

## Test plan
- [x] `tests/unit/stores/history.spec.ts` covers both orders (stream already running when the GET starts, and the turn finishing while the GET is in flight) plus a load that starts after the turn and must still replace.
- [x] `make -C frontend test` (2139 passed) and `make -C frontend lint`
- [ ] CI chromium shard that runs `admin-impersonation-chat.spec.ts` — the recorded failure was `message-done` never becoming visible after a 200 `/messages/stream`


orgaralf (approved):
LGTM

